PR #3, opened 2022-03-17
Update bug report and feature templates

IanVS:
I found that references to Trivago were still in a few places, and we're not using discussions here, so I tweaked the templates slightly.


blutorange (approved):
Looks good, I don't see any spelling mistakes. I did a quick grep on the repo, there's one more references to trivago in the package.json

> https://github.com/IanVS/prettier-plugin-sort-imports/blob/documentation-updates/package.json#L32

I suppose we should keep that though, since this is a fork? But you might want to add yourself.

The CHANGELOG.md has some more references to issues over at https://github.com/trivago/prettier-plugin-sort-imports, but I guess that's fine too.

PR #6, opened 2022-04-18
Preserve directive prologues fixes #5 #7

### LINKED ISSUE #5: 'use strict' pushed below imports
**Your Environment**
 * **Prettier version**: 2.6.2
 * **node version**: 16.14.0
 * **package manager**: 8.3.1
 * **IDE**: CLI

**Describe the bug**

`'use strict'` is pushed below imports.

**To Reproduce**

1. Try to format this code:
```typescript
'use strict'

import type { Period } from './Period'
```
2. Bug appears.
```typescript
import type { Period } from './Period'

;('use strict')
```

**Expected behavior**

Keep `'use strict'` above.

```json
{
  "semi": false,
  "trailingComma": "none",
  "singleQuote": true,
  "printWidth": 640,
  "importOrder": ["@riot/metaplayer", "^[./]"],
  "importOrderSeparation": true
}
```

**Error log**

**Contribute to @trivago/prettier-plugin-sort-imports**

- [ ] I'm willing to fix this bug 🥇

blutorange:
This fixes #5 and also includes the fix for #7 since I need it for this.  While the relevance of strict directives may be doubtful for modules, correctness is still a goal and we should still preserve the original code and only order imports. 

For references, this code

```javascript
// below is a directive prologue
  'use custom' ; /* more directives... */ 'enable typecheck'
                 'forbid IE'

import './SetupEnvironment';
import type { Period } from './Period'

"this is not a directive prologue";

function foo() {
}
```

now becomes

```javascript
// below is a directive prologue
"use custom";
/* more directives... */

"enable typecheck";
"forbid IE";

import "./SetupEnvironment";

import type { Period } from "./Period";

("this is not a directive prologue");

function foo() {}
```

and the following

```javascript
; 'use strict'

import './SetupEnvironment';
import type { Period } from './Period'

"this is not a directive prologue";

function foo() {
}
```

still becomes

```javascript
import "./SetupEnvironment";

import type { Period } from "./Period";

("use strict");

("this is not a directive prologue");

function foo() {}
```

since an empty statement is not a directive prologue according [to the spec](https://tc39.es/ecma262/#sec-directive-prologues-and-the-use-strict-directive).

### DIFF
--- a/src/utils/get-code-from-ast.ts
+++ b/src/utils/get-code-from-ast.ts
@@ -22,10 +22,12 @@ export const getCodeFromAst = (
     interpreter?: InterpreterDirective | null,
 ) => {
     const allCommentsFromImports = getAllCommentsFromNodes(nodes);
+    const allCommentsFromDirectives = getAllCommentsFromNodes(directives);
 
     const nodesToRemoveFromCode = [
         ...nodes,
         ...allCommentsFromImports,
+        ...allCommentsFromDirectives,
         ...(interpreter ? [interpreter] : []),
         ...directives,
     ];
--- a/src/utils/get-code-from-ast.ts
+++ b/src/utils/get-code-from-ast.ts
@@ -1,5 +1,5 @@
 import generate from '@babel/generator';
-import { InterpreterDirective, Statement, file } from '@babel/types';
+import { Directive, InterpreterDirective, Statement, file } from '@babel/types';
 
 import { newLineCharacters } from '../constants';
 import { getAllCommentsFromNodes } from './get-all-comments-from-nodes';
@@ -7,12 +7,18 @@ import { removeNodesFromOriginalCode } from './remove-nodes-from-original-code';
 
 /**
  * This function generate a code string from the passed nodes.
- * @param nodes all imports
- * @param originalCode
+ * @param nodes All imports, in the sorted order in which they should appear in
+ * the generated code.
+ * @param originalCode The original input code that was passed to this plugin.
+ * @param directives All directive prologues from the original code (e.g.
+ * `"use strict";`).
+ * @param interpreter Optional interpreter directives, if present (e.g.
+ * `#!/bin/node`).
  */
 export const getCodeFromAst = (
     nodes: Statement[],
     originalCode: string,
+    directives: Directive[],
     interpreter?: InterpreterDirective | null,
 ) => {
     const allCommentsFromImports = getAllCommentsFromNodes(nodes);
@@ -21,6 +27,7 @@ export const getCodeFromAst = (
         ...nodes,
         ...allCommentsFromImports,
         ...(interpreter ? [interpreter] : []),
+        ...directives,
     ];
 
     const codeWithoutImportsAndInterpreter = removeNodesFromOriginalCode(
@@ -31,7 +38,7 @@ export const getCodeFromAst = (
     const newAST = file({
         type: 'Program',
         body: nodes,
-        directives: [],
+        directives: directives,
         sourceType: 'module',
         interpreter: interpreter,
         sourceFile: '',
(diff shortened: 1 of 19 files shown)


blutorange:
note to self: there's still an issue with duplicated comments

blutorange:
(the above is expected to fail, I am not done yet)

IanVS:
I can't give this a thorough review right now, but I wonder, is there a way to check if there is a performance impact to this change?

blutorange:
We could of course do it manually, not sure if we can add performance tests. But regarding how it "should" affect the performance:

* Preserving directive prologues should not affect it all all, since I'm only using the result from the parse that's already being done
* Other than that, I only changed how the imports are removed from the code. Before, a Regex was created for each import, then run in the code. Now, I build a list of ranges to replace (which calls sort on the list of ranges, but that's around `O(n log n)` and the input size is only the number of imports present in the file), do a bunch of substrings to extract the parts to keep, then join those parts back together.

So I wouldn't expect this to have an effect on performance (if anything it could improve it due to less Regexes). But perhaps we should take a large JS file and compare the times it takes to format, just so we can rule out an unexpected bottleneck.

blutorange:
@IanVS Do you have time to take a look at this in the next few days? Otherwise do you mind if I do a quick test myself and then merge?

IanVS:
I'll give it a quick glance through today and approve.  Curious about perf, but I don't think that's a blocker.

IanVS:
@blutorange In reading the spec, I see that it says:

>A [Directive Prologue](https://tc39.es/ecma262/#directive-prologue) may be an empty sequence.

I'm not so familiar with the spec, is an empty sequence different from an empty statement?  I guess an empty sequence is `""`?

src/utils/get-code-from-ast.ts
```diff
@@ -1,26 +1,35 @@
...
  * This function generate a code string from the passed nodes.
- * @param nodes all imports
- * @param originalCode
+ * @param nodes All imports, in the sorted order in which they should appear in
+ * the generated code.
+ * @param originalCode The original input code that was passed to this plugin.
+ * @param directives All directive prologues from the original code (e.g.
+ * `"use strict";`).
+ * @param interpreter Optional interpreter directives, if present (e.g.
+ * `#!/bin/node`).
```
IanVS, line 16: These comments are helpful, thanks.

src/utils/remove-nodes-from-original-code.ts
```diff
@@ -1,20 +1,165 @@
...
+
+    const resultRanges: Range[] = [];
+    const isValidRange = (start: number, end: number) => end > start;
+
+    // Add the part from the start of the enclosing range to the start of the
+    // first range to exclude.
+    //
+    // |-----------xxxxx-----xxxx-----xxxx-----------|
+    //  ^---------^
+    //  This part
```
IanVS, line 121: 😍

blutorange:
@IanVS  As far as I understand it

> A Directive Prologue is the longest sequence of ExpressionStatements [...] where each ExpressionStatement in the sequence consists entirely of a StringLiteral token followed by a semicolon.

An empty sequence is like the empty set, so that's just their way of saying that having the directive prologue is optional.

This is a source file with an empty directive prologue list, consisting of an empty statement and a function declaration statement.
```javascript
;
function foo() {}
```

This is a source file with a single-element directive sequence, and the value of the directive is the empty string.
```javascript
"";
function foo() {}
```

This is also source file with a single-element directive sequence. (The semicolon is inserted via the process of automatic semicolon insertion)
```javascript
""
function foo() {}
```

It doesn't really matter though, since we aren't doing the parsing ourselves and just rely on the parser being implemented correctly. I just added some tests for these cases just in case somebody gets the idea to implement the parsing manually ; )

PS: You can use e.g. https://astexplorer.net/ to see how a piece of code parses.

IanVS (approved):
I tested this out in my own project, and it seems to make no difference in performance, prettier takes roughly 14.5 seconds to check the formatting either way, and I did not get any formatting problems when running this branch.  Nicely done!

PR #10, opened 2022-05-11
Introduce `importOrderBuiltinModulesToTop` for Node Builtins

IanVS:
This is a cherry-pick of a PR from @nickhudkins to the upstream project: https://github.com/trivago/prettier-plugin-sort-imports/pull/150, adjusted slightly to work within the rule of this project to not sort beyond side-effect import groups.  

@blutorange what do you think of this?  I'm reluctant to add another option, but if we changed this without an option it would be a breaking change.  I don't have a strong opinion one way or the other, but personally I do like sorting builtins to the top, so it would be nice to have some way to do it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -213,6 +213,16 @@ with options as a JSON string of the plugin array:
 importOrderParserPlugins: []
 ```
 
+
+#### `importOrderBuiltinModulesToTop`
+
+**type**: `boolean`
+
+**default value:** `false`
+
+A boolean value to enable sorting of builtins to the top of all import groups.
+
+
 ### How does import sort work ?
 
 The plugin extracts the imports which are defined in `importOrder`. These imports are considered as _local imports_.
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -1,5 +1,6 @@
 import { ParserPlugin } from '@babel/parser';
 import { expressionStatement, stringLiteral } from '@babel/types';
+import { builtinModules } from 'module';
 
 export const flow: ParserPlugin = 'flow';
 export const typescript: ParserPlugin = 'typescript';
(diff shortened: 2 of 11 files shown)


blutorange:
The commit itself looks good, it's only a very small change.

@IanVS Have you checked if this works for prefixed modules such as `import {...} from "node:fs"` or `import * as fs from "node:fs/promises"`?

Other than that, my very personal opinion is that a prettier plugin should be opinionated in the same way prettier is and have as few options as possible. But this plugin doesn't really follow that philosophy. I took a quick look at https://www.npmjs.com/package/@ianvs/prettier-plugin-sort-imports and I'd say there are enough downloads we can't just introduce a breaking change.

In the long term we should decide if we want to tend towards "many options for configurability" or "few options with an opinionated sort order" -- we could remove options in a major release. What are you thoughts about this? I also wonder what users are thinking in this regard.

But in short, if we add this without a new major release, it needs to be an option that defaults to false. And a major release imo only makes sense if we decided to set a new goal for this plugin and start removing other option.

IanVS:
Thanks @blutorange.  I've added some more tests as you suggested.  

I agree with you about not making breaking changes outside of majors.  It annoys me that prettier itself makes breaking changes in minors, and I don't want to follow that pattern.

I think one thing we should consider is that this plugin is changing the actual AST, rather than only formatting like prettier does.  So, there's a chance we do need to allow a few more options than a purely-formatting tool would.  But where it makes sense, I think we should avoid options.  I'd like to make this behavior the default in the next major version, and maybe even remove the option entirely, unless we hear that it causes big problems for folks.  

I think this is ready for another review, when you have time.  Thanks again!

blutorange:
> I'd like to make this behavior the default in the next major version, and maybe even remove the option entirely, unless we hear that it causes big problems for folks.

Semver only works if people actually stick to it, so yes, sounds like a good idea to me

I also don't really like the `importOrder` option -- it requires me to manually add the dependency I use. At the least I'd prefer a sensible default behavior when no option is set that might perhaps group all imports with the same scope automatically. But that's outside the scope of this issue.

IanVS:
I'm going to merge this so I can rebase https://github.com/IanVS/prettier-plugin-sort-imports/pull/12 on main.  I think I got an approval above, but if you want any changes, @blutorange, let me know and I'll open another PR.

PR #12, opened 2022-05-12
Custom import order separation

### LINKED ISSUE #11: [Feature] Allow manual configuration of line breaks
See https://github.com/trivago/prettier-plugin-sort-imports/issues/36

I think the solution in https://github.com/trivago/prettier-plugin-sort-imports/pull/42 is pretty smart.  It does not require a new option, but gives some flexibility for folks who want to specify new lines between specific groups.

We could wait until the upstream project merges this, but it seems a little unlikely at this point.

@blutorange, what do you think of this approach?

IanVS:
Closes #11

This is an adaptation of https://github.com/trivago/prettier-plugin-sort-imports/pull/42 from @atombrenner.  I wasn't able to use most of the commits from the original branch, because our approach is different in this project, but I did pull in his commit with tests.  

The other change I needed to make here was to modify the cloned nodes in `adjustCommentsOnSortedNodes`, rather than mutating the original, which was leading to some strange bugs in the tests.

### DIFF
--- a/src/utils/adjust-comments-on-sorted-nodes.ts
+++ b/src/utils/adjust-comments-on-sorted-nodes.ts
@@ -9,31 +9,30 @@ import { ImportOrLine } from '../types';
  * they were in the original nodes.
  * @param nodes A list of nodes in the order as they were originally.
  * @param finalNodes The same set of nodes, but in the final sorting order.
+ * @returns A copied and adjusted set of nodes, containing comments
  */
 export const adjustCommentsOnSortedNodes = (
     nodes: ImportDeclaration[],
     finalNodes: ImportOrLine[],
 ) => {
-    // maintain a copy of the nodes to extract comments from
+    // We will mutate a copy of the finalNodes, and extract comments from the original
     const finalNodesClone = finalNodes.map(clone);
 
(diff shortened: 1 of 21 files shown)


src/utils/adjust-comments-on-sorted-nodes.ts
```diff
@@ -9,31 +9,30 @@ import { ImportOrLine } from '../types';
...
  * @param nodes A list of nodes in the order as they were originally.
  * @param finalNodes The same set of nodes, but in the final sorting order.
+ * @returns A copied and adjusted set of nodes, containing comments
  */
 export const adjustCommentsOnSortedNodes = (
     nodes: ImportDeclaration[],
     finalNodes: ImportOrLine[],
 ) => {
-    // maintain a copy of the nodes to extract comments from
+    // We will mutate a copy of the finalNodes, and extract comments from the original
```
fbartho, line 18: Great clarification

fbartho (approved):
Your example test cases look great, and your documentation changes look like a good fit.

I see that the tests are executed in CI, so I don't have any objections to this PR :)

IanVS:
@blutorange I'm going to merge this, but let me know if you want any changes and I can make them in a follow-up.

PR #15, opened 2022-05-17
Fix: "Ignore unknown option importOrderBuiltinModulesToTop"

fbartho:
I pulled release 3.4.0 and added the new rule, and unfortunately got these complaints from prettier:

```
[warn] Ignored unknown option { importOrderBuiltinModulesToTop: true }.
```

I compared other Config Options, and this is my guess at the necessary fix

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -46,6 +46,12 @@ const options = {
         default: false,
         description: 'Should specifiers be sorted?',
     },
+    importOrderBuiltinModulesToTop: {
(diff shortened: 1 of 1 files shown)


fbartho:
./cc @IanVS sorry I didn't find this before your release!

IanVS (approved):
Whoops, yeah good catch, thanks.  Not sure how I missed this either.  We might need to take a look at our tests, and maybe add some cli tests.

PR #17, opened 2022-05-17
More strict on what is considered a builtin - Fixes #16

fbartho:
Relates to #16

Looks like a combination of config options are conflicting?

Not sure why it triggers on CLI but not in IDE for me, but the run_spec is able to reproduce it in this repo.

PR #18, opened 2022-05-17
Remove node 12 tests, add node 18

IanVS:
Node 12 has reached end of life, and node 18 is the `current` version: https://nodejs.org/en/about/releases/

I don't think this requires a breaking change just yet, but testing with node 12 is causing issues with our tests in https://github.com/IanVS/prettier-plugin-sort-imports/pull/17.


fbartho (approved):

fbartho:
Merging to make #17 go green. If necessary we can always restore node12 in the matrix!

PR #19, opened 2022-05-18
Feature: `importOrderMergeDuplicateImports`

### LINKED ISSUE #4: [Feature] Combine type and value imports
Not so much a problem, but an enhancement

I'm not positive this fits in to the scope of this project, but thought I'd suggest it anyway.  I recently created https://github.com/IanVS/type-import-codemod, which combines type and value imports into a single import statement, using a new feature of Typescript 4.5: `import {foo, type Foo} from './foo';`.  If desired, I think it could be adapted or used within this prettier plugin.  

If we did this, it would probably either need to be behind an option, or ideally, the version of typescript would be checked to find out whether it supports the new syntax, perhaps with an option to disable it.  I normally want to avoid options, but since this feature isn't directly related to sorting imports, there should be a way to disable it, if we even include the functionality at all.  

I could see creating a different prettier plugin just to combine those imports.  That's honestly probably the better approach than trying to shim it into this one.

**Additional context**
I'm opening this issue to see if anyone else has thoughts about whether this would be useful.

fbartho:
This PR implements import declaration merging for duplicate imports.

- I had to adjust `getCodeFromAST` so that we could still provide the original nodes, even if some of the nodes were being deleted.
- It leverages the existing chunking logic, so if there's a `side-effect` import, or a `prettier-ignore` import, it won't merge across that boundary.
- I did not implement conversion of `import type { stuff }` into `import { type stuff }` as it was not obvious when this would be wise. If we want to control that with an additional config option, something like `importOrderMergeTypeStyle: "none" | "to-normal-import" | "unmerge"` I could be convinced. I forsee people wanting opposite sides of the coin for that one.

Please let me know what you think and if there's any edge-cases I missed!

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,7 +86,7 @@ yarn add --dev @ianvs/prettier-plugin-sort-imports
 
 ## Usage
 
-Add an order in prettier config file.
+Add your preferred settings in your prettier config file.
 
 ```javascript
 module.exports = {
@@ -96,10 +96,16 @@ module.exports = {
   "singleQuote": true,
   "semi": true,
   "importOrder": ["^@core/(.*)$", "^@server/(.*)$", "^@ui/(.*)$", "^[./]"],
+  "importOrderBuiltinModulesToTop": true,
+  "importOrderCaseInsensitive": true,
+  "importOrderParserPlugins": ["typescript", "jsx", "decorators-legacy"],
+  "importOrderMergeDuplicateImports": true,
   "importOrderSeparation": true,
-  "importOrderSortSpecifiers": true
+  "importOrderSortSpecifiers": true,
 }
 ```
+
+_Note: all flags are off by default, so explore your options [below](#apis)_
 
 ### APIs
 
@@ -203,6 +209,14 @@ compared with `"importOrderCaseInsensitive": true`:
 import ExamplesList from './ExamplesList';
 import ExampleView from './ExampleView';
 ```
+
+#### `importOrderMergeDuplicateImports`
+
+**type**: `boolean`
+
+**default value:** `false`
+
+A boolean value to enable or disable multiple import statements referencing the same source. Not all patterns can be merged! Notably: `import type …` will not be converted to `import {type …` or vice-versa.
 
 #### `importOrderParserPlugins`
 
--- a/README.md
+++ b/README.md
@@ -216,7 +216,7 @@ import ExampleView from './ExampleView';
 
 **default value:** `false`
 
-A boolean value to enable or disable multiple import statements referencing the same source. Not all patterns can be merged! Notably: `import type …` will not be converted to `import {type …` or vice-versa.
+When `true`, multiple import statements from the same module will be combined into a single import.
 
 #### `importOrderParserPlugins`
 
--- a/README.md
+++ b/README.md
@@ -88,7 +88,7 @@ yarn add --dev @ianvs/prettier-plugin-sort-imports
 
 Add an order in prettier config file.
 
-```ecmascript 6
+```javascript
 module.exports = {
   "printWidth": 80,
   "tabWidth": 4,
--- a/src/preprocessor.ts
+++ b/src/preprocessor.ts
@@ -11,11 +11,11 @@ export function preprocessor(code: string, options: PrettierOptions): string {
     const {
         importOrderParserPlugins,
         importOrder,
-        importOrderCaseInsensitive,
-        importOrderSeparation,
-        importOrderGroupNamespaceSpecifiers,
-        importOrderSortSpecifiers,
         importOrderBuiltinModulesToTop,
+        importOrderCaseInsensitive,
+        importOrderGroupNamespaceSpecifiers,
+        importOrderSeparation,
+        importOrderSortSpecifiers,
     } = options;
 
     const importNodes: ImportDeclaration[] = [];
@@ -40,19 +40,25 @@ export function preprocessor(code: string, options: PrettierOptions): string {
         },
     });
 
-    // short-circuit if there are no import declaration
+    // short-circuit if there are no import declarations
     if (importNodes.length === 0) {
         return code;
     }
 
-    const allImports = getSortedNodes(importNodes, {
+    const remainingImports = getSortedNodes(importNodes, {
         importOrder,
-        importOrderCaseInsensitive,
-        importOrderSeparation,
-        importOrderGroupNamespaceSpecifiers,
-        importOrderSortSpecifiers,
         importOrderBuiltinModulesToTop,
+        importOrderCaseInsensitive,
+        importOrderGroupNamespaceSpecifiers,
+        importOrderSeparation,
+        importOrderSortSpecifiers,
     });
 
-    return getCodeFromAst(allImports, code, directives, interpreter);
+    return getCodeFromAst({
+        nodes: remainingImports,
+        importNodes,
+        originalCode: code,
+        directives,
+        interpreter,
+    });
 }
--- a/src/preprocessor.ts
+++ b/src/preprocessor.ts
@@ -19,7 +19,7 @@ export function preprocessor(code: string, options: PrettierOptions): string {
         importOrderSortSpecifiers,
     } = options;
 
-    const importNodes: ImportDeclaration[] = [];
+    const allOriginalImportNodes: ImportDeclaration[] = [];
     const parserOptions: ParserOptions = {
         sourceType: 'module',
         plugins: getExperimentalParserPlugins(importOrderParserPlugins),
@@ -36,17 +36,17 @@ export function preprocessor(code: string, options: PrettierOptions): string {
                 isTSModuleDeclaration(p),
             );
             if (!tsModuleParent) {
-                importNodes.push(path.node);
+                allOriginalImportNodes.push(path.node);
             }
         },
     });
 
     // short-circuit if there are no import declarations
-    if (importNodes.length === 0) {
+    if (allOriginalImportNodes.length === 0) {
         return code;
     }
 
-    const remainingImports = getSortedNodes(importNodes, {
+    const nodesToOutput = getSortedNodes(allOriginalImportNodes, {
         importOrder,
         importOrderBuiltinModulesToTop,
         importOrderCaseInsensitive,
@@ -57,8 +57,8 @@ export function preprocessor(code: string, options: PrettierOptions): string {
     });
 
     return getCodeFromAst({
-        nodes: remainingImports,
-        importNodes,
+        nodesToOutput,
+        allOriginalImportNodes,
         originalCode: code,
         directives,
         interpreter,
--- a/src/preprocessor.ts
+++ b/src/preprocessor.ts
@@ -14,6 +14,7 @@ export function preprocessor(code: string, options: PrettierOptions): string {
         importOrderBuiltinModulesToTop,
         importOrderCaseInsensitive,
         importOrderGroupNamespaceSpecifiers,
+        importOrderMergeDuplicateImports,
         importOrderSeparation,
         importOrderSortSpecifiers,
     } = options;
@@ -50,6 +51,7 @@ export function preprocessor(code: string, options: PrettierOptions): string {
         importOrderBuiltinModulesToTop,
         importOrderCaseInsensitive,
         importOrderGroupNamespaceSpecifiers,
+        importOrderMergeDuplicateImports,
         importOrderSeparation,
         importOrderSortSpecifiers,
     });
--- a/src/types.ts
+++ b/src/types.ts
@@ -6,6 +6,7 @@ export interface PrettierOptions extends RequiredOptions {
     importOrderCaseInsensitive: boolean;
     importOrderBuiltinModulesToTop: boolean;
     importOrderGroupNamespaceSpecifiers: boolean;
+    importOrderMergeDuplicateImports: boolean;
     importOrderSeparation: boolean;
     importOrderSortSpecifiers: boolean;
     // should be of type ParserPlugin from '@babel/parser' but prettier does not support nested arrays in options
@@ -28,9 +29,16 @@ export type GetSortedNodes = (
         | 'importOrderBuiltinModulesToTop'
         | 'importOrderCaseInsensitive'
         | 'importOrderGroupNamespaceSpecifiers'
+        | 'importOrderMergeDuplicateImports'
         | 'importOrderSeparation'
         | 'importOrderSortSpecifiers'
     >,
 ) => ImportOrLine[];
 
 export type GetChunkTypeOfNode = (node: ImportDeclaration) => string;
+
+export type GetImportFlavorOfNode = (node: ImportDeclaration) => string;
+
+export type MergeNodesWithMatchingImportFlavors = (
+    nodes: ImportDeclaration[],
+) => ImportDeclaration[];
--- a/src/types.ts
+++ b/src/types.ts
@@ -1,6 +1,15 @@
 import { ExpressionStatement, ImportDeclaration } from '@babel/types';
 import { RequiredOptions } from 'prettier';
 
+import {
+    chunkTypeOther,
+    chunkTypeUnsortable,
+    importFlavorIgnore,
+    importFlavorSideEffect,
+    importFlavorType,
+    importFlavorValue,
+} from './constants';
+
 export interface PrettierOptions extends RequiredOptions {
     importOrder: string[];
     importOrderCaseInsensitive: boolean;
@@ -13,9 +22,16 @@ export interface PrettierOptions extends RequiredOptions {
     importOrderParserPlugins: string[];
 }
 
+export type ChunkType = typeof chunkTypeOther | typeof chunkTypeUnsortable;
+export type FlavorType =
+    | typeof importFlavorIgnore
+    | typeof importFlavorSideEffect
+    | typeof importFlavorType
+    | typeof importFlavorValue;
+
 export interface ImportChunk {
     nodes: ImportDeclaration[];
-    type: string;
+    type: ChunkType;
 }
 
 export type ImportGroups = Record<string, ImportDeclaration[]>;
@@ -35,9 +51,9 @@ export type GetSortedNodes = (
     >,
 ) => ImportOrLine[];
 
-export type GetChunkTypeOfNode = (node: ImportDeclaration) => string;
+export type GetChunkTypeOfNode = (node: ImportDeclaration) => ChunkType;
 
-export type GetImportFlavorOfNode = (node: ImportDeclaration) => string;
+export type GetImportFlavorOfNode = (node: ImportDeclaration) => FlavorType;
 
 export type MergeNodesWithMatchingImportFlavors = (
     nodes: ImportDeclaration[],
--- a/src/types.ts
+++ b/src/types.ts
@@ -5,11 +5,11 @@ export interface PrettierOptions extends RequiredOptions {
     importOrder: string[];
     importOrderCaseInsensitive: boolean;
     importOrderBuiltinModulesToTop: boolean;
+    importOrderGroupNamespaceSpecifiers: boolean;
+    importOrderSeparation: boolean;
+    importOrderSortSpecifiers: boolean;
     // should be of type ParserPlugin from '@babel/parser' but prettier does not support nested arrays in options
     importOrderParserPlugins: string[];
-    importOrderSeparation: boolean;
-    importOrderGroupNamespaceSpecifiers: boolean;
-    importOrderSortSpecifiers: boolean;
 }
 
 export interface ImportChunk {
@@ -27,8 +27,8 @@ export type GetSortedNodes = (
         | 'importOrder'
         | 'importOrderBuiltinModulesToTop'
         | 'importOrderCaseInsensitive'
-        | 'importOrderSeparation'
         | 'importOrderGroupNamespaceSpecifiers'
+        | 'importOrderSeparation'
         | 'importOrderSortSpecifiers'
     >,
 ) => ImportOrLine[];
--- a/src/utils/__tests__/get-code-from-ast.spec.ts
+++ b/src/utils/__tests__/get-code-from-ast.spec.ts
@@ -47,13 +47,16 @@ it('merges duplicate imports correctly', () => {
 // second comment
 import z from 'z';
 import c from 'c';
+import type {C} from 'c';
+import type {See} from 'c';
 import g from 'g';
 import t from 't';
 import k from 'k';
 import a from 'a';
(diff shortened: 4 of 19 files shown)


README.md
```diff
@@ -204,6 +210,14 @@ import ExamplesList from './ExamplesList';
 import ExampleView from './ExampleView';
 ```
 
+#### `importOrderMergeDuplicateImports`
```
IanVS, line 213: Can you explain your thoughts on why this should be an option?  @blutorange and I both believe that this project already has too many options, and we'd like to be more opinionated and rely less on having to set options to get reasonable results.  I think we shouldn't be afraid to make breaking changes and releasing major versions, if needed.

IanVS (commented):
I don't have time for a full review right now, but thanks for hopping in and working on this.  I did have one question that would be good to discuss, and I'll try to give a full review soon.

src/utils/__tests__/merge-nodes-with-matching-flavors.spec.ts
```diff
@@ -0,0 +1,162 @@
...
+        // * as J, * as K can't merge because both Namespaces
+        import { I } from \\"c\\";
+        import { default as Def1, default as Def2 } from \\"d\\";
+        import Foo1 from \\"e\\";
+        import Foo2 from \\"e\\";
+        import { Junk2 } from \\"junk-group-2\\";
+        "
+    `);
+});
+it("doesn't merge duplicate imports if option disabled", () => {
```
blutorange, line 83: Tests are always appreciated, it's great that you added some. But regarding snapshot tests, there's a special directory for those: https://github.com/IanVS/prettier-plugin-sort-imports/tree/main/tests that uses Jest snapshot testing. Could you move these tests there? That would keep it a bit more consistent.

blutorange: Snapshots  tests are integration tests that check whether the plugin as a whole works as expected. https://github.com/IanVS/prettier-plugin-sort-imports/tree/main/src/utils/__tests__ is meant for unit tests of the single functions in the source code. And while having unit tests won't hurt, personally I agree with you and consider integration tests that check how the plugin works as a whole more important for this plugin.

src/utils/merge-nodes-with-matching-flavors.ts
```diff
@@ -0,0 +1,166 @@
...
+    nodeToForget: ImportDeclaration,
+) {
+    if (!mergeIsSafe(nodeToKeep, nodeToForget)) {
+        return false;
+    }
+
+    nodeToKeep.specifiers.push(...nodeToForget.specifiers);
+
+    // The line numbers will be all messed up. Is this a problem?
+    nodeToKeep.leadingComments = [
```
blutorange, line 97: I'm wondering what exactly people would expect how comments are merged. Consider

```js
// comment 1
// comment 2
import { foo } from "a"
// comment 3
// comment 4
// comment 5
// comment 6
import { bar } from "a"
// comment 7
// comment 8
```

which results in

```js
// comment 1
// comment 2
// comment 3
// comment 4
// comment 5
// comment 6
import { foo, bar  } from "a"
// comment 7
// comment 8
```

If one deletes the second import manually and adds the import to the first, one would get

```js
// comment 1
// comment 2
import { foo, bar } from "a"
// comment 3
// comment 4
// comment 5
// comment 6
// comment 7
// comment 8
```

I can see an argument for both cases. Though I'd assume comments in the import section are rare in general and if they do occur, they are probably meant as an explanation for a particular import, so keeping comments on top sound like the right thing to do.

src/utils/merge-nodes-with-matching-flavors.ts
```diff
@@ -0,0 +1,166 @@
...
+    nodeToKeep: ImportDeclaration,
+    nodeToForget: ImportDeclaration,
+) {
+    if (!mergeIsSafe(nodeToKeep, nodeToForget)) {
+        return false;
+    }
+
+    nodeToKeep.specifiers.push(...nodeToForget.specifiers);
+
+    // The line numbers will be all messed up. Is this a problem?
```
blutorange, line 96: [There's the part](https://github.com/IanVS/prettier-plugin-sort-imports/commit/b5c23c79174682452ccd46ec2feddc49448880a5#diff-ce50d77c8138cb15e5b03ef96053bc615ad9450e5c30ada045bf72de79d6966eR164-R195) where it removes nodes from the original code which relies on the line number being correct (it did so even before the linked commit). 

It is called from https://github.com/IanVS/prettier-plugin-sort-imports/blob/b5c23c79174682452ccd46ec2feddc49448880a5/src/utils/get-code-from-ast.ts#L35

I think that might be alright since it should operate on the original nodes, but perhaps you could check again that the nodes passed to that function are not the modified ones?

src/utils/__tests__/merge-nodes-with-matching-flavors.spec.ts
fbartho: @blutorange I guess I'm confused as to the exact ask.

I'm definitely asserting specific functionality of the merge-nodes-with-matching-flavors, I just found it easier to validate the output by calling through the higher level API, because the output is just source-code, rather than somehow creating "artificial" arrays of nodes and shoving them into just the merge-function itself.

In this next PR: https://github.com/IanVS/prettier-plugin-sort-imports/pull/20 I added a bunch of additional test cases to the same file -- those additional test cases [came from here](https://github.com/IanVS/type-import-codemod/blob/9f3210861a77216f07317537a04ce61467e0df7e/src/transform.test.ts) and those were indeed unit tests as well.

I think one cause of the organizational comments is that I'm using `toMatchInlineSnapshot` for developer-updating convenience, rather than `toEqual` as seen in the source-tests which requires the developer to update any changes by hand.

Options I see:
A. Rewrite tests to target lower level API `mergeNodesWithMatchingImportFlavors` directly
B. Rewrite tests to use `toEqual` as seen in other test cases and the source-tests
C. Rewrite tests to `toMatchInlineSnapshot -> toMatchSnapshot` + to move these tests to the special directory for traditional snapshots

What would you two like me to do?

src/utils/merge-nodes-with-matching-flavors.ts
IanVS: I think that the behavior shown as current (all comments go above) is the "more correct" option.  If that's what the code is doing now, I say we stick with it.

src/utils/merge-nodes-with-matching-flavors.ts
fbartho: The nodes passed do both the modified & the original ones -- exactly line: https://github.com/IanVS/prettier-plugin-sort-imports/pull/19/commits/cf4e6af3093740dd1b035c3d798ae8ddb7af10ca#diff-ab5d98b73e57fb0fe6cb1b2b9ea7d14530182d5d1987eae7c545bd15f4528cfeR37

The trick is that the modified nodes don't have their line-numbers adjusted. The line-numbers still match what's in the provided code.

So… I think we're good?

src/utils/merge-nodes-with-matching-flavors.ts
fbartho: I think the current behavior is reasonable, but maybe unintuitive. I think the root cause is that the parser prefers to assign comments to the `leading` position than the `trailing` position?

So comments 3, 4, 5, 6 are `leading` comments for the 2nd import, and my code moves them to be `leading` comments for the 1st import.

If comments 3-6 were considered `trailing` comments for the 1st import, then the behavior would have matched the `manual` process and maybe been a hair more intuitive?

Again, I agree the way it behaves now is probably the "more correct" option even if it doesn't match how a human would visualize the operations. Leading vs trailing comments is a weird thing to specify tbh.

src/utils/merge-nodes-with-matching-flavors.ts
blutorange: Yes, I think so. If not, and if it did remove code at the wrong position, then the tests would have failed already

src/utils/merge-nodes-with-matching-flavors.ts
IanVS: /shrug it's intuitive to me, if we're combining the comments above each of the imports to remain above the combined import.  I also think this is a super rare edge case, and probably not worth spending too much effort on perfecting.  ;)

src/utils/__tests__/merge-nodes-with-matching-flavors.spec.ts
fbartho: I have now pushed that change of `toMatchInlineSnapshot -> toEqual` to both PRs #19, #20

src/utils/__tests__/get-import-flavor-of-node.spec.ts
```diff
@@ -0,0 +1,32 @@
...
+import { g } from "g";
+`,
+            { plugins: ['typescript'] },
+        ).map((node) => getImportFlavorOfNode(node)),
+    ).toMatchInlineSnapshot(`
+        Array [
+          "side-effect",
+          "regular",
+          "type",
+          "regular",
```
IanVS, line 27: Hmmmm, this one feels odd.  This comes from `import { type C } from "c";`, which seems like a type import, but I guess even in astexplorer, this would be considered an importDeclaration with `value` importKind, containing an importSpecifier with `type` importKind.  So yeah, this is correct, just funky.

src/utils/get-import-flavor-of-node.ts
```diff
@@ -0,0 +1,29 @@
...
+
+/**
+ * Classifies nodes by import-flavor, primarily informing whether the node is a candidate for merging
+ *
+ * @param node
+ * @returns {("prettier-ignore"|"regular"|"side-effect"|"type")}
+ */
+export const getImportFlavorOfNode: GetImportFlavorOfNode = (node) => {
+    const hasIgnoreNextNode = (node.leadingComments ?? []).some(
+        (comment) => comment.value.trim() === 'prettier-ignore',
```
IanVS, line 17: Question, will this count as an ignored node?

```js
// prettier-ignore
// something else
import foo from 'foo';
```

Should we only consider the comment directly above the import?  Honest question, I'm not sure what the right answer is.

src/utils/get-import-flavor-of-node.ts
```diff
@@ -0,0 +1,29 @@
...
+    importFlavorSideEffect,
+    importFlavorType,
+} from '../constants';
+import type { GetImportFlavorOfNode } from '../types';
+
+/**
+ * Classifies nodes by import-flavor, primarily informing whether the node is a candidate for merging
+ *
+ * @param node
+ * @returns {("prettier-ignore"|"regular"|"side-effect"|"type")}
```
IanVS, line 13: Using jsdoc together with typescript is a bit tricky.  I'd say this belongs as a return type in typescript on the function below, and we should use `@returns` to describe the return with a short phrase or sentence.  In this case, we could say `@returns the flavor of the import node`, or something like that.

src/utils/get-import-flavor-of-node.ts
```diff
@@ -0,0 +1,29 @@
...
+ */
+export const getImportFlavorOfNode: GetImportFlavorOfNode = (node) => {
```
IanVS, line 15: It seems like some of this is duplicated logic with `getChunkTypeOfNode`, perhaps it can be consolidated?

src/types.ts
```diff
@@ -27,10 +28,17 @@ export type GetSortedNodes = (
...
         | 'importOrderGroupNamespaceSpecifiers'
+        | 'importOrderMergeDuplicateImports'
+        | 'importOrderSeparation'
         | 'importOrderSortSpecifiers'
     >,
 ) => ImportOrLine[];
 
 export type GetChunkTypeOfNode = (node: ImportDeclaration) => string;
+
+export type GetImportFlavorOfNode = (node: ImportDeclaration) => string;
```
IanVS, line 40: This type is only used in the function declaration itself.  We can remove this, and specify the types inline at the function declaration itself.

src/utils/__tests__/get-code-from-ast.spec.ts
```diff
@@ -36,3 +41,43 @@ import z from "z";
...
+    const formatted = getCodeFromAst({
+        nodes: sortedNodes,
+        importNodes,
+        originalCode: code,
+        directives: [],
+    });
+    expect(format(formatted, { parser: 'babel' })).toEqual(
+        `// first comment
+// second comment
+import a, { b } from "a";
```
IanVS, line 75: 👍 good test, checking default + named export combination.

src/utils/__tests__/get-sorted-nodes-by-import-order.spec.ts
```diff
@@ -187,11 +190,12 @@ test('it returns all sorted nodes with sort order case-insensitive', () => {
     const result = getImportNodes(code);
     const sorted = getSortedNodesByImportOrder(result, {
         importOrder: ['^a$', '^t$', '^k$', '^B', '^[./]'],
+        importOrderBuiltinModulesToTop: false,
         importOrderCaseInsensitive: true,
-        importOrderSeparation: false,
         importOrderGroupNamespaceSpecifiers: false,
+        importOrderMergeDuplicateImports: false,
+        importOrderSeparation: false,
```
IanVS, line 197: I prefer not making formatting-only changes like alphabetizing as part of a large PR like this, it adds noise.  Let's make small tweaks like this (especially when there's a lot of them) in a separate PR.  But don't undo it here.

src/utils/__tests__/merge-nodes-with-matching-flavors.spec.ts
```diff
@@ -0,0 +1,160 @@
...
+    import { G } from 'b';
+    import * as J from 'c';
+    import { Junk2 } from 'junk-group-2'
+    import * as K from "c";
+    // * as J, * as K can't merge because both Namespaces
+    import {I} from "c"
+    import { default as Def2 } from 'd';
+    import { default as Def1 } from 'd';
+    import Foo1 from 'e';
+    import Foo2 from 'e';
```
IanVS, line 34: There's a ton going on here, it might make the test more readable if it were broken into separate test cases.

src/types.ts
IanVS: I guess there are a lot of those in this file, we can clean that up later.  But I think that `string` could be a bit more specific, and we could use the union of the values we actually expect to get, here.

src/utils/merge-nodes-with-matching-flavors.ts
```diff
@@ -0,0 +1,166 @@
...
+        (groups, node) => {
+            const flavor = getImportFlavorOfNode(node);
+            if (isMergeableFlavor(flavor)) {
+                groups[flavor].push(node);
+            }
+            return groups;
+        },
+        {
+            [importFlavorRegular]: [] as ImportDeclaration[],
+            [importFlavorType]: [] as ImportDeclaration[],
```
IanVS, line 35: Typecasting works and is probably fine here, but normally I prefer to use a generic on the `reduce` call, like 

```ts
return nodes.reduce<{[key: typeof importFlavorRegular | typeof importFlavorType]: ImportDeclaration[]}>(
```

That's a bit clunky in this case, but it lets us avoid using `as`.  Go ahead and leave this as-is, though, just wanted to mention it.

src/utils/merge-nodes-with-matching-flavors.ts
```diff
@@ -0,0 +1,166 @@
...
+}
+
+/**
+ * Modifies context, deleteContext,
+ * case A: context has no node for an import source, then it's assigned.
+ * case B: context has a node for an import source, then it's merged, and old node is added to deleteContext.
+ */
+function mutateContextAndMerge({
+    context,
+    deleteContext,
```
IanVS, line 120: Might as well call this `nodesToDelete`, which is pretty clear, and also the name of the variable used when calling this function.

src/utils/merge-nodes-with-matching-flavors.ts
```diff
@@ -0,0 +1,166 @@
...
+    node: ImportSpecifier | ImportDefaultSpecifier | ImportNamespaceSpecifier,
+): node is ImportDefaultSpecifier {
+    return node.type === 'ImportDefaultSpecifier';
+}
+
+/** Return false if the merge will produce an invalid result */
+function mergeIsSafe(
+    nodeToKeep: ImportDeclaration,
+    nodeToForget: ImportDeclaration,
+) {
```
IanVS, line 59: These are good to check for!  :+1:  And thanks for the comments.

src/utils/merge-nodes-with-matching-flavors.ts
IanVS: I don't quite understand this, and it relates to my comment from earlier about why we're providing both the modified and original nodes.  But, feel free to ignore me, if y'all think it's fine how it is.

IanVS (commented):
Overall I think this looks good.  I had a few minor comments and questions, but I think this is pretty close to mergable.  Thanks for the work on this, @fbartho.

src/preprocessor.ts
```diff
@@ -40,19 +41,26 @@ export function preprocessor(code: string, options: PrettierOptions): string {
...
     });
 
-    // short-circuit if there are no import declaration
+    // short-circuit if there are no import declarations
     if (importNodes.length === 0) {
         return code;
     }
 
-    const allImports = getSortedNodes(importNodes, {
+    const remainingImports = getSortedNodes(importNodes, {
```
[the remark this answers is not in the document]
fbartho: Yes, the return value of `getSortedNodes` is no longer "all imports" it's just the "imports that weren't deleted"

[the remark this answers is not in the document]
fbartho: I'll rename this inline with renames I'm doing to match [your feedback on naming of parameters to `getCodeFromAst`](https://github.com/IanVS/prettier-plugin-sort-imports/pull/19#discussion_r877641457)

src/utils/get-code-from-ast.ts
```diff
@@ -9,23 +9,32 @@ import { removeNodesFromOriginalCode } from './remove-nodes-from-original-code';
...
  */
-export const getCodeFromAst = (
-    nodes: Statement[],
-    originalCode: string,
-    directives: Directive[],
-    interpreter?: InterpreterDirective | null,
-) => {
+export const getCodeFromAst = ({
+    nodes,
+    importNodes = nodes,
```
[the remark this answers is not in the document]
fbartho: Yes, `importNodes` is a superset, we need to pass the original nodes so that the old code-locations can be deleted when we merge a node. Once I was deleting some of the nodes from the `nodes`-array then the generated output code had the properly sorted & modified nodes followed by the unmodified nodes I wanted to have deleted.

How do you feel about these names?

```ts
/**
 * This function generates a code string from the passed nodes.
 * @param nodesToOutput The remaining imports which should be rendered. (Node specifiers & types may be mutated)
 * @param allOriginalImportNodes All import nodes that were originally relevant. (This includes nodes that need to be deleted!)
 * @param originalCode The original input code that was passed to this plugin.
 * @param directives All directive prologues from the original code (e.g.
 * `"use strict";`).
 * @param interpreter Optional interpreter directives, if present (e.g.
 * `#!/bin/node`).
 */
export const getCodeFromAst = ({
    nodesToOutput,
    allOriginalImportNodes = nodesToOutput,
    originalCode,
    directives,
    interpreter,
}: {
    nodesToOutput: Statement[];
    allOriginalImportNodes?: Statement[];
    originalCode: string;
    directives: Directive[];
    interpreter?: InterpreterDirective | null;
}) => {
```

src/utils/__tests__/get-import-flavor-of-node.spec.ts
```diff
@@ -0,0 +1,32 @@
+import { getImportFlavorOfNode } from '../get-import-flavor-of-node';
+import { getImportNodes } from '../get-import-nodes';
+
+it('should correctly classify a bunch of import expressions', () => {
```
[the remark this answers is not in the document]
fbartho: A bunch of the individual cases for this are actually added in #20 -- would you mind reviewing the test file in there instead?

This is just a bulk test that I had originally, lazily created. If after reviewing the tests there you still want me to delete or recreate this test, please let me know!

src/utils/get-import-flavor-of-node.ts
fbartho: I copied this helper method from `getChunkTypeOfNode` which this method was heavily inspired by.

I assumed that that was the "intentional" logic here, but if that logic is wrong. I'm totally fine hoisting both of them somewhere and changing it somehow.

Ref: https://github.com/IanVS/prettier-plugin-sort-imports/blob/3b72a60b254d8fab9e8c362b31be54d028309e30/src/utils/get-chunk-type-of-node.ts#L23-L25

src/types.ts
fbartho: I was only trying to match the patterns I saw already in the codebase. But happy to make the return type more explicit!

Doing that in a commit now. -- Technically this is a typescript-breaking change but not a runtime change.

src/utils/__tests__/merge-nodes-with-matching-flavors.spec.ts
fbartho: More granular unit tests were submitted in #20

src/utils/merge-nodes-with-matching-flavors.ts
fbartho: That comment was a bit out of date, as I've now learned that that's actually an important attribute of how `getCodeFromAst` functions. It needs the old line-numbers/loc to properly move things around!

```ts
// These mutations don't update the line numbers, and that's crucial for moving things around.
// To get updated line-numbers you would need to re-parse the code after these changes are rendered!
```

src/utils/get-import-flavor-of-node.ts
fbartho: (I've now hoisted this out for sharing between the two locations)

src/utils/merge-nodes-with-matching-flavors.ts
fbartho: Did it, but slightly differently!

fbartho:
@IanVS PR Feedback applied to this and #20 -- happy to do another round when you get another chance to look.

src/types.ts
IanVS: Yep, I'm pretty new to actually working in this code myself, so some of the patterns that exist here are a little odd.  Not blaming you, but any means.  :-D. Best to stay consistent for now, and then cleanup later if we decide to.

IanVS (approved):
Great work here, I pushed a small commit to add a couple of tests to convince myself everything's working correctly, and a few small comments to remind myself what is going on when I look at this again in the future.

PR #20, opened 2022-05-18
Feature: `importOrderMergeTypeImportsIntoRegular`

fbartho:
Fixes: #4

Allows the ability for users to convert `import type` expressions into `import {type …}` expressions via a boolean flag.

Should this be controlled via a string parameter?

### DIFF
--- a/src/utils/merge-nodes-with-matching-flavors.ts
+++ b/src/utils/merge-nodes-with-matching-flavors.ts
@@ -5,6 +5,7 @@ import type {
     ImportNamespaceSpecifier,
     ImportSpecifier,
 } from '@babel/types';
+import assert from 'assert';
 
 import {
     importFlavorRegular,
@@ -51,6 +52,25 @@ function nodeIsImportDefaultSpecifier(
 ): node is ImportDefaultSpecifier {
     return node.type === 'ImportDefaultSpecifier';
 }
+function nodeIsImportSpecifier(
+    node: ImportSpecifier | ImportDefaultSpecifier | ImportNamespaceSpecifier,
+): node is ImportSpecifier {
+    return node.type === 'ImportSpecifier';
+}
+
+function convertImportSpecifierType(node: ImportSpecifier) {
+    assert(node.importKind === 'value' || node.importKind === 'type');
+    node.importKind = 'type';
+}
+
+/** Pushes an `import type` expression into `import { type …}` */
+function convertTypeImportToValueImport(node: ImportDeclaration) {
+    assert(node.importKind === 'type');
+    node.importKind = 'value';
+    node.specifiers
+        .filter(nodeIsImportSpecifier)
+        .forEach(convertImportSpecifierType);
+}
 
 /** Return false if the merge will produce an invalid result */
 function mergeIsSafe(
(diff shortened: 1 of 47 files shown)


IanVS:
I think we need to talk about whether this warrants yet another option, but in the meantime, do you think you could add some tests?  You might be able to use the test cases from my codemod as a starting point: https://github.com/IanVS/type-import-codemod/blob/9f3210861a77216f07317537a04ce61467e0df7e/src/transform.test.ts

fbartho:
Happy to add those tests. I do tend to have macro-tests that assert a ton of functionality all at once. Not sure that's a good sign, but code-coverage likes it 😉

fbartho:
@IanVS Test cases imported, the only differences I can see
- this plugin sorts the output by source, while the codemod didn't
- this the codemod grouped `type` imports together, while this plugin just sorts all the fields by name.

Do we need/want to group type imports when sorting?

IanVS:
Sorry, it looks like there are a number of conflicts on this PR now.  Your best bet might be to reset hard to main, and then cherry-pick the commits from this PR back onto the branch.

fbartho:
@IanVS sorry about the lag re: updates, I was under the weather last week! `main` is now merged in. Most conflicts were trivially solved with a 3-headed merge (using p4merge), and the few that weren't automatically solved don't seem too bad.

Please take another look, and let me know if this can proceed.

src/utils/merge-nodes-with-matching-flavors.ts
```diff
@@ -51,19 +51,35 @@ function selectNodeImportSource(node: ImportDeclaration) {
...
 ): node is ImportDefaultSpecifier {
     return node.type === 'ImportDefaultSpecifier';
 }
+function nodeIsImportSpecifier(
+    node: ImportSpecifier | ImportDefaultSpecifier | ImportNamespaceSpecifier,
+): node is ImportSpecifier {
+    return node.type === 'ImportSpecifier';
+}
+
+function convertImportSpecifierType(node: ImportSpecifier) {
```
IanVS, line 70: Can we call this `convertImportSpecifierToType`?

IanVS (commented):
It took me a while to get to this, sorry about that.  I think it mostly looks good.  I had a few small comments/questions.  

Unless you or @blutorange feels strongly, I'd like to sort the type imports to the end.  I think it looks better that way, and is more readable.  Would you mind making that change?  Then we can merge and release this, and prepare for the breaking changes release.

IanVS:
Hi @fbartho.  How would you like to proceed on this one? We can merge it as-is if you don't have time to make changes, and I can play around with re-ordering the type/value import specifiers in a separate PR.

fbartho:
Hiya! I was traveling 2 weeks ago, and totally missed these messages!

No objections to the changes, I’m happy to make them — should be able to get to them this week or this coming Saturday

IanVS:
Great, thanks!  I hope you had a good time on your travels!

IanVS:
Hi @fbartho just checking in to see if this is still something you'll be able to get to.  NBD if not, just was wondering.

fbartho:
@IanVS whoops! It's been a little bit of a scramble. I'll take a look at this, this week!

IanVS:
Thanks for your work on this @fbartho!  I pushed up a few tweaks, and am going to merge / release.  We can continue to iterate, but I'd like to start preparing for a breaking release, to include prettier 3 support.

IanVS:
I ran this on my codebase, and found that this was incorrectly merged:

From: 
```ts
import { ComponentProps, useEffect } from 'react';
import type React from 'react';
```

To: 
```ts
import { ComponentProps, useEffect, React } from 'react';
```

I'm looking into it now.

PR #21, opened 2022-05-18
Maintenance: use fine-grained lodash-imports

fbartho:
Lodash is a huge and chunky module. If we directly specify which methods we need from it, we can avoid a bit of the always-keep-it-up-to-date treadmill.

Further investigation could be done to see if we really even need the `clone`/`isEqual` methods from lodash, but at least with this PR change we put a soft fence around it to keep from panicking if lodash has security incidents. And to keep it from spreading 😉.

Not a big deal if you don't want to drop it, since the version of whatwg-url that is used by jsdom used by jest still needs it (even if newer versions don't), but I thought I'd provide a PR.

I'm aware that in lodash v5 they claim they will stop deploying the individual module packages, but that argument is based on "runtime package bundling" arguments, rather than on the maintenance concerns.

### DIFF
--- a/package.json
+++ b/package.json
@@ -50,11 +50,13 @@
         "@babel/traverse": "^7.17.3",
         "@babel/types": "^7.17.0",
         "javascript-natural-sort": "0.7.1",
-        "lodash": "^4.17.21"
+        "lodash.clone": "^4.5.0",
+        "lodash.isequal": "^4.5.0"
     },
     "devDependencies": {
         "@types/jest": "^27.4.1",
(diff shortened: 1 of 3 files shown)


blutorange:
That's definitely a good change. As we've seen in the recent years, more dependencies are not always better. `isEqual(nodes[0].loc, node.loc)` could probably be replaced with an explicit comparison between the properties of the location (which shouldn't be many), which would only leave the `clone` import.

fbartho:
Okay, merging this PR for now, and will propose a future PR to strip the lodash imports entirely.

Issue #23, opened 2022-05-19
Check tie-breaking with importOrderCaseInsensitive


IanVS:
As mentioned in https://github.com/IanVS/prettier-plugin-sort-imports/issues/22#issuecomment-1132090026, we should check to be sure we correctly break ties when two imports differ only in capitalization, and `importOrderCaseInsensitive` is `true`.  Maybe this is already handled, I haven't checked.

blutorange:
While we're at it, we also need to break ties in case two imports only differ by whether they have a type specifier. I could make this a separate issue, but since it's both related to tie-breaking, I'd like to keep it in one issue

https://github.com/trivago/prettier-plugin-sort-imports/issues/116

fbartho:
I don't think this is a super big issue if we can count on a stable sort algorithm. If we can't then we have trouble. (Not-to mention that case-sensitivity issues are only possible on some-file-systems or with fake packages in node's resolution hierarchy.)

NodeJS's sort is not guaranteed stable, so I recommend we switch to a stable-sort. Here's a random one from npm: https://www.npmjs.com/package/stable / https://github.com/Two-Screen/stable -- it has zero deps, and does have TypeScript.

Additionally, I learned that `a.localeCompare(b, 'en', {numeric: true, ignorePunctuation: true})` can be used for Natural Language sorting from [a blog](https://fuzzytolerance.info/blog/2019/07/19/The-better-way-to-do-natural-sort-in-JavaScript/)

So this proposal adds a dep, in exchange for removing a dep & a hand-written type-definition file. haha

fbartho:
I learned today that Javascript's language spec now requires `Array.prototype.sort` be a Stable-sort (since 2019 on V8!). https://v8.dev/features/stable-sort

Not sure when this made it to NodeJS, but I'm sure that implies we don't even need a dependency at all. That means this proposal allows us to remove the dependency & handwritten type-definition file and just use built-in JS! 🎉

fbartho:
Opened a related PR in #28

IanVS:
Hi @fbartho, I'm not exactly sure that having a stable sort is what we want in this case.  As @blutorange mentioned in the comment in the issue description:

>we need to make sure that we break ties by falling back to case-sensitive sorting

Just because a sort is stable, doesn't mean this will happen, it means that whatever order the rows were in to begin with is how they will stay, so 

```
import { a } from "someTHING"; 
import { a } from "SOMEthing"; 
```

[...]

Issue #22, opened 2022-05-19
New major version -- what do we want to change


blutorange:
Regarding #4, now might be a good a time as any to start a new major release and get rid of some of the legacy weight. Especially now that the number of users is still rather low. Perhaps we can start by brainstorming what we would like to change.

As a general guideline, I also don't think we need to make a plugin that can be customized to suit for everybody's code style. For one thing, there are other plugins such as  @trivago/prettier-plugin-sort-imports.

And, more importantly, prettier itself is already an opinionated code formatter that prescribes a certain code style.  For reference, here's is their rationale: https://prettier.io/docs/en/option-philosophy.html

![image](https://user-images.githubusercontent.com/2456413/169220822-a8b8915f-dc93-4014-b187-308c7fc08f9c.png)

They only want options that are required for some purpose, such as to support external tools. But no options that are just people's opinions on how the code style should look like.

Also: people who need a very particular code style won't be using prettier, and in extension, they also wouldn't use this plugin. I beliebe prettier plugins should keep at least somewhat to the spirit of prettier. Having a single plugin that by now probably has accumulated more options than the program for which the plugin is meant feels a bit strange. Having many options also makes tests and maintenance a pain, since you now have to test various combinations of options, or get lots of bug reports if you don't.

## Let's review the options we've got

* importOrderBuiltinModulesToTop - We wanted to remove this in the next major release, it always puts builtins at the top.
* importOrderParserPlugins - Needs to stay, this is required for the plugin to work correctly
* importOrderSortSpecifiers - Any reason not to remove this and have it enabled always?
* importOrderSeparation - Are people really that picky about having new lines? I also work with Java + Eclipse and it just groups imports and separates them by new lines, and nobody in our team has an issue with this. I'd just remove this and have it  
* importOrderGroupNamespaceSpecifiers - I'm honestly not exactly sure what this does. I also don't see any tests where this is set to `true`?
* importOrderCaseInsensitive - This is an option I could image we might want to keep. Any opinions? If we don't, what would be the value for this?
* importOrder - Other than the technical `importOrderParserPlugins`, I feel like this should be the main option of this plugin, if not the only one.
  * The default for this option is `[]`, which I feel is a pretty bad default. It just sorts everything alphabetically, including relative imports etc. 
  * The other thing I don't like about this option is that it requires you to specify each `@scope` manually. Adding a new dependency now requires you to review and sometimes change the prettier config for this plugin.
[...]

PR #28, opened 2022-05-29
Feature: drop javascript-natural-sort dep

fbartho:
Minor problem: localeCompare appears to always interleave capital & lowercase letters.

I don't know how to do both numeric-sort AND case-sensitive (classic `Array.sort()` mode) at the same time using pure locale-compare.
This is how I was forced to replicate the original case-sensitive sort: https://github.com/IanVS/prettier-plugin-sort-imports/blob/5f07e195c2e411109aab25de39109a94acfdd492/src/natural-sort/index.ts#L14-L16

Relates to: https://github.com/IanVS/prettier-plugin-sort-imports/issues/23#issuecomment-1140533424

### DIFF
--- a/package.json
+++ b/package.json
@@ -49,7 +49,6 @@
         "@babel/parser": "^7.17.7",
         "@babel/traverse": "^7.17.3",
         "@babel/types": "^7.17.0",
-        "javascript-natural-sort": "0.7.1",
         "lodash.clone": "^4.5.0",
         "lodash.isequal": "^4.5.0"
     },
--- a/src/natural-sort/__tests__/natural-sort.spec.ts
+++ b/src/natural-sort/__tests__/natural-sort.spec.ts
@@ -1,0 +1,54 @@
+import { naturalSort } from '..';
+
+it('should sort normal things alphabetically', () => {
+    expect(
+        ['a', 'h', 'b', 'i', 'c', 'd', 'j', 'e', 'k', 'f', 'g'].sort((a, b) =>
+            naturalSort(a, b, { importOrderCaseInsensitive: false }),
+        ),
+    ).toEqual(['a', 'b', 'c', 'd', 'e', 'f', 'g', 'h', 'i', 'j', 'k']);
+});
+
+it('should treat capital letters as alphabetically earlier than lowercase if case sensitive (or unspecified)', () => {
+    expect(
+        ['./ExampleView', './ExamplesList'].sort((a, b) =>
+            naturalSort(a, b, { importOrderCaseInsensitive: false }),
+        ),
+    ).toMatchInlineSnapshot(
+        ['./ExampleView', './ExamplesList'],
+        `
+        Array [
+          "./ExampleView",
+          "./ExamplesList",
+        ]
+    `,
+    );
+    expect(
+        ['./ExampleView', './ExamplesList'].sort((a, b) =>
+            naturalSort(a, b, {}),
+        ),
+    ).toMatchInlineSnapshot(
+        ['./ExampleView', './ExamplesList'],
+        `
+        Array [
+          "./ExampleView",
+          "./ExamplesList",
+        ]
+    `,
+    );
+});
+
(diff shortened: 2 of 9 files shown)


fbartho:
Note: the minor issue goes away if we delete the `importOrderCaseInsensitive` option as discussed in #22

Then we just always sort case-insensitive, and numerically! 🎉 🍰

IanVS (commented):
In general I'm in favor of relying on native JS apis when possible, but I have some concern that the behavior here could be different from javascript-natural-sort, and we could introduce some unintentional breaking changes.  

I think we can use the relevant test cases from https://github.com/Bill4Time/javascript-natural-sort/blob/master/unit-tests.html to minimize that chance, but since I don't think this fixes https://github.com/IanVS/prettier-plugin-sort-imports/issues/23, and is really just a dependency cleanup, let's hold off on merging it until we're ready to release v4, which I think will be relatively soon.  What do you think?

fbartho:
@IanVS — note: either #23 was already fine, because modern NodeJS uses modern V8 which has Stable Sort now, or JavaScript-natural-sort *wasn’t stable*, in which case this ticket **does fix** #23.

Within any *chunk*, tie breaking should have a deterministic outcome once when you use a stable sorting algorithm with a defined comparator (edge cases exist, but don’t apply here).

So… either this ticket fixes #23, or #23 is moot, but either way, I support shipping this with no-options in the breaking release. (Caveat locale decision).

IanVS:
I don't think that stable sorting fixes issue #23.  For example, if you add this test case to this branch's `natural-sort.spec.ts`, it will fail:

```
it('should sort deterministically if case-insensitive', () => {
    expect(
        ['./something', './Something'].sort((a, b) =>
            naturalSort(a, b, { importOrderCaseInsensitive: true }),
        ),
    ).toEqual(['./something', './Something']);

    expect(
        ['./Something', './something'].sort((a, b) =>
            naturalSort(a, b, { importOrderCaseInsensitive: true }),
        ),
    ).toEqual(['./something', './Something']);
});
```

The order in which the imports were _originally_ defined determines what order they are placed in (which is what it means to have a stable sort).  This is _not_ what we want, however.  We want a _deterministic_ sort.  Given two imports, in any order, they should always have the same final order.

fbartho:
>Maybe we are not agreeing on what it means to be case insensitive. To me, it means, don't worry about whether capitals come before lowercase, or vise versa. So, Z-dep shouldn't come before a-dep. But, given A-dep and a-dep, we need an answer for which should come first, since in a case-insensitive sort, they're exactly the same.

Ah! That indeed is our difference. `localeCompare` in insensitive mode says `(a-dep, A-dep) = 0`, `(A-dep, a-dep) = 0` so either order is identical. To me that's internally consistent with "insensitive", so that's what I've been visualizing.

IanVS:
Given that we've cleared up the confusion over https://github.com/IanVS/prettier-plugin-sort-imports/issues/23, I think we can close out this PR.  But, let me know if you disagree and we can discuss further.  Thanks for your work on this, and sorry for the miscommunication.

IanVS:
Given that you and @blutorange believe the best option is to avoid dictating the order of imports differing only by case sensitivity, I think it's worth re-visiting this.

fbartho:
@IanVS I pushed an update to this  that made it compatible with `next` as of 9am, but lol, I see there's already more conflicts. I'll wait to hear if you want more from me on this or if you want to take it over!

IanVS:
Whoops, sorry I hadn't seen that you pushed that.  Let's finish the option cleanup and then come back to this afterwards, if that's cool with you?

fbartho:
However you prefer works for me. Was just trying to be helpful :P

For what it's worth, I think it's just a trivial conflict if you want to fold this PR in and forget about it!

IanVS (approved):
Thanks for bearing with me on this one.  Glad we can finally get this cleanup in.

PR #29, opened 2022-05-30
Avoid initial new line when using custom group separator

ArnaudBarre:
I think this screenshot will highlight what I feel like an inconsistency:

Screenshot 2022-05-30 at 15 15 22

### DIFF
--- a/src/utils/get-sorted-nodes-by-import-order.ts
+++ b/src/utils/get-sorted-nodes-by-import-order.ts
@@ -70,8 +70,13 @@ export const getSortedNodesByImportOrder: GetSortedNodes = (nodes, options) => {
     for (const group of importOrder) {
         // If it's a custom separator, all we need to do is add a newline
         if (isCustomGroupSeparator(group)) {
+            const lastNode = finalNodes[finalNodes.length - 1];
+            // Avoid empty new line if first group is empty
+            if (!lastNode) {
+                continue;
+            }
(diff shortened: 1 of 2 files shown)


fbartho (approved):
This feels like a good fix to me, I was noticing similar issues when I applied this to a large project. Thoughts @IanVS @blutorange?

src/utils/get-sorted-nodes-by-import-order.ts
```diff
@@ -70,8 +70,13 @@ export const getSortedNodesByImportOrder: GetSortedNodes = (nodes, options) => {
     for (const group of importOrder) {
         // If it's a custom separator, all we need to do is add a newline
         if (isCustomGroupSeparator(group)) {
+            const lastNode = finalNodes[finalNodes.length - 1];
+            // Avoid empty new line if first group is empty
+            if (!lastNode) {
```
IanVS, line 75: This feels a little bit indirect.  Could we check if `finalNodes.length === 0`? Would that accomplish the same thing?

ArnaudBarre: It would be the same thing for this first checks, but then I would need to assert TS that the list is non empty for the next check (or add a useless runtime check)

IanVS:
Thanks for the contribution, @ArnaudBarre!

PR #38, opened 2022-10-24
Fix: Do not combine default type imports with value imports

IanVS:
This fixes a bug which was causing default type imports to be combined into value imports, which isn't safe to do.  For instance,

```ts
import { ComponentProps, useEffect } from 'react';
import type React from 'react';
```

was turning into

```ts
import { ComponentProps, useEffect, React } from 'react';
```

Which is not correct.  This change prevents default type imports from ever being merged.

### DIFF
--- a/README.md
+++ b/README.md
@@ -228,7 +228,17 @@ _Note:_ If you want greater control over which groups are separated from others,
 
 **default value:** `false`
 
-A boolean value to enable or disable sorting of the specifiers in an import declarations.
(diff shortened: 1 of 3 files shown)


IanVS:
I'm going to merge this to get the fix released, but I'd still love an _ex post facto_ review if anyone has some time.

fbartho (commented):
Looks great @IanVS, good catch!

PR #39, opened 2022-10-25
Fix: broken lodash import

mipyykko:
I think another lodash import was added after refactoring to use more fine-grained lodash import in https://github.com/IanVS/prettier-plugin-sort-imports/pull/21


IanVS:
Thanks for the catch!

PR #43, opened 2022-10-25
Upgrade to Prettier 3 in dev dependencies

IanVS:
This is a bit of a test to see whether we're compatible with prettier 3.  Initially we were broken, due to https://github.com/prettier/prettier/issues/13729.  But that's been addressed in the latest alpha, and tests are now passing.  

I think we could merge and release this prior to the next breaking change.  The only user-facing change in this PR is to peerDependencies, so that folks can start using prettier 3 if they want.  But we might as well hold off, I think, to see how things shake out in prettier, since it sounds like there may be more breaking changes on the way.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,7 +5,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        node: [14.x, 16.x, 18.x]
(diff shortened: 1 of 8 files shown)


IanVS:
Replaced by https://github.com/IanVS/prettier-plugin-sort-imports/pull/75

Issue #35, opened 2022-09-04
Type imports


mattiaz9:
**Is your feature request related to a problem?**

I'd like to sort all my types into one single group.

Example:
```ts
import { a } from 'foo'
import { b } from 'bar'

import type { A } form 'foo'
import type { B } form 'bar'
```

**Describe the solution you'd like**

Something like this:
https://github.com/trivago/prettier-plugin-sort-imports/pull/153

**Describe alternatives you've considered**

None

IanVS:
@mattiaz9, do you prefer that over this syntax?

```ts
import { a, type A } from 'foo';
import { b, type B } from 'bar';
```

mattiaz9:
@IanVS yes, I prefer to separate them

tusharf5:
@mattiaz9 you can try https://github.com/serverless-guru/prettier-import-order which is a fork of this project and have this feature.

mattiaz9:
Thank you @tusharf5, but I can't put the types where I want, so I rather wait for the feature to be implemented

mattiaz9:
@IanVS I'm trying out the new version, but it doesn't seem to work. 
I think it might going in conflict with other imports. For instance `"<TYPES>^[./]"` is in conflict with `"^[./]"`.

Or maybe I'm doing something wrong. This is my config:
```js
module.exports = {
  importOrder: [
    "^react",
    "^next",
    "<THIRD_PARTY_MODULES>",
    "",
    ".s?css$",
    "^@heroicons",
    "^@/assets",
    "",
    "^[./]",
    "^@/",
    "",
    "<TYPES>^[./]",
    "<TYPES>^react",
    "<TYPES>^next",
    "<TYPES>^@/",
    "<TYPES>",
  ],
  importOrderSeparation: false,
  importOrderMergeDuplicateImports: true,
}
```

This is an example result of how it sorts the imports:
```ts
import React, {  useCallback,  useEffect,  useState } from "react"
import isHotkey from "is-hotkey"
import { createEditor, Transforms, Editor as SlateEditor } from "slate"
import { Slate, Editable, ReactEditor } from "slate-react"

import Block from "./blocks/Block"
import type { BlockProps } from "./blocks/Block"
import Field from "./blocks/Field"
import Image from "./blocks/Image"
import type { LeafProps } from "./blocks/Leaf"
import Leaf from "./blocks/Leaf"
import Page from "./blocks/Page"
import type { PageProps } from "./blocks/Page"
[...]

PR #44, opened 2022-10-26
Support explicit type sorting

IanVS:
Closes https://github.com/IanVS/prettier-plugin-sort-imports/issues/35

Adapted from https://github.com/trivago/prettier-plugin-sort-imports/pull/153, by @Xenfo.  

This adds a new special string, `<TYPES>` that can be added to the `importOrder` array.  When used, it will cause type imports to be sorted as specified by its location in the array.  

Notes:
- If it is used, it will disable `importOrderCombineTypeAndValueImports`, throwing a warning if both are used.  This is because:
- We will split apart type and value import declarations if `<TYPES>` is used, so that types can be sorted appropriately.  
- Thinking towards the next breaking change when we remove options, I think the default will be to enable `importOrderCombineTypeAndValueImports`, and this change will give users a good way to opt-out of that behavior if they want, by specifying the location for `<TYPES>`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -200,6 +200,12 @@ To move the third party imports at desired place, you can use `<THIRD_PARTY_MODU
 "importOrder": ["^@core/(.*)$", "<THIRD_PARTY_MODULES>", "^@server/(.*)$", "^@ui/(.*)$", "^[./]"],
 ```
 
+If you would like to order type imports differently from value imports, you can use the special `<TYPES>` string.  This example will place third party types at the top, followed by local types, then third party value imports, and lastly local value imports:
+
+```json
+"importOrder": ["<TYPES>", "<TYPES>^[./]", "<THIRD_PARTY_MODULES>", "^[./]"],
+```
+
 #### `importOrderSeparation`
 
 **type**: `boolean`
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -26,8 +26,9 @@ export const mergeableImportFlavors = [
  * Used to mark the position between RegExps,
  * where the not matched imports should be placed
  */
-export const THIRD_PARTY_MODULES_SPECIAL_WORD = '<THIRD_PARTY_MODULES>';
 export const BUILTIN_MODULES = `^(?:node:)?(?:${builtinModules.join('|')})$`;
+export const THIRD_PARTY_MODULES_SPECIAL_WORD = '<THIRD_PARTY_MODULES>';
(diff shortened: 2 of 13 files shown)


IanVS:
@fbartho @blutorange I'd love to get your thoughts on this change before merging.

README.md
```diff
@@ -200,6 +200,12 @@ To move the third party imports at desired place, you can use `<THIRD_PARTY_MODU
 "importOrder": ["^@core/(.*)$", "<THIRD_PARTY_MODULES>", "^@server/(.*)$", "^@ui/(.*)$", "^[./]"],
 ```
 
+If you would like to order type imports differently from value imports, you can use the special `<TYPES>` string.  This example will place third party types at the top, followed by local types, then third party value imports, and lastly local value imports:
+
+```json
+"importOrder": ["<TYPES>", "<TYPES>^[./]", "<THIRD_PARTY_MODULES>", "^[./]"],
```
fbartho, line 206: I don't love the "virtual matcher" combined with a custom matcher for Types. `"<TYPES>^[./]"`
It feels like a bad precedent to have to split these keys into tags and a regex, and have to match them by `startsWith`, etc.

I don't mind the rest of this PR though! We talked about this feature previously.

IanVS: I agree it does feel a bit wonky.  Do you have any idea how else we could support it, though?  (We don't actually use `startsWith` here, fwiw, but we do remove the `<TYPES>` part when creating the regex.  

The tricky part here is giving the flexibility to declare a regex as only applying to type imports vs value imports.  I think we'd need to support objects in `importOrder`, and use something like `importOrder: [{regex: '^[./]', isType: true}, ...]`, and that doesn't feel great either.  I'm open to suggestions here, though.

fbartho: Okay, so thinking about this more deeply, this is supposed to mirror the ability to create arbitrary groups for types. Eg. interleave types and non-type imports in batches.

I'd mayhaps try to argue that as a plugin for an opinionated formatter, maybe we should fight to remove the `importOrder` option entirely, but that seems like a hard sell.

Fewer customization points = simpler code, and easier onboarding for users, so I guess your original proposal is better than the alternatives that jump to mind.

IanVS: Yeah I think that it's okay to keep `importOrder` and try to minimize the number of other options that we expose.  I definitely think we need a default for it that works in most cases, though.

Issue #56, opened 2023-03-24
importOrderCombineTypeAndValueImports: true throws error for specific type syntax for flow projects


andidev:
**Your Environment**

-   **Prettier version**: 2.7.1
-   **node version**: v16.17.0
-   **package manager**: npm@9.3.1
-   **IDE**: Webstorm

**Describe the bug**
This only seem to happens for flow projects.
Sorting imports with `type` prefix with option `importOrderCombineTypeAndValueImports: true` throws error.
Not sure maybe I am missing something setting here but I think I confugured prettier babel plugin correctly to prase flow files.

<img width="961" alt="image" src="https://user-images.githubusercontent.com/571389/227526017-50da3eb4-a2c2-4cbe-808d-fc7cb90b1cfb.png">

is causing following error:
<img width="1872" alt="image" src="https://user-images.githubusercontent.com/571389/227526421-49a9fb3d-07b5-4551-8622-6c6f4c7cc2c3.png">

**Expected behavior**
No error is expected since its valid flow syntax

**Configuration File (cat .prettierrc, prettier.config.js, .prettier.js)**

```
/**
 * @prettier
 */
module.exports = {
    arrowParens: 'avoid',
    printWidth: 200,
    tabWidth: 4,
    semi: true,
    singleQuote: true,
    jsxSingleQuote: true,
    trailingComma: 'es5',
    requirePragma: true,
    bracketSpacing: true,
    parser: 'flow',
    quoteProps: 'as-needed',
    overrides: [
        {
            files: 'src/constants/*.js',
            options: {
                quoteProps: 'consistent',
            },
        },
        {
            files: 'src/i18n/*.js',
            options: {
                printWidth: 10000,
                quoteProps: 'consistent',
            },
        },
    ],
    importOrder: ['<THIRD_PARTY_MODULES>', '^rootpath$', '^(scripts/.*)$', '(^src/.*)$', '^(test/.*)$'],
    importOrderSeparation: false,
    importOrderSortSpecifiers: true,
    importOrderBuiltinModulesToTop: true,
    importOrderMergeDuplicateImports: true,
    importOrderCombineTypeAndValueImports: true,
    importOrderParserPlugins: ['flow', 'jsx'],
};
```

**Contribute to @ianvs/prettier-plugin-sort-imports**

-   [ ] I'm willing to fix this bug 🥇

IanVS:
[...]

PR #57, opened 2023-03-24
Fix importOrderCombineTypeAndValueImports for Flow

IanVS:
Fixes https://github.com/IanVS/prettier-plugin-sort-imports/issues/56

I needed to loosen the assertion a bit, because Flow's AST does not apparently use an `importKind` of `"value"` for specifiers, instead it is `null`.

I also added an e2e test here, for both Flow and TypeScript.

### DIFF
--- a/src/utils/merge-nodes-with-matching-flavors.ts
+++ b/src/utils/merge-nodes-with-matching-flavors.ts
@@ -76,7 +76,12 @@ function nodeIsImportSpecifier(
 }
 
 function convertImportSpecifierToType(node: ImportSpecifier) {
-    assert(node.importKind === 'value' || node.importKind === 'type');
+    assert(
+        node.importKind === 'value' ||
+            node.importKind === 'type' ||
+            // importKind can be null when using Flow
+            node.importKind === null,
+    );
     node.importKind = 'type';
 }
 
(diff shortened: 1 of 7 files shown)


IanVS:
@andidev, if you'd like to test this out in your project, you can find this package in your node_modules and update `convertImportSpecifierToType` in `lib/src/utils/merge-nodes-with-matching-flavors.js` to:

```js
function convertImportSpecifierToType(node) {
    (0, assert_1.default)(node.importKind === 'value' ||
        node.importKind === 'type' ||
        // importKind can be null when using Flow
        node.importKind === null);
    node.importKind = 'type';
}
```

fbartho (approved):
I’m not a flow expert or user really, so I defer to you on this one. This change seems relatively low sensitivity, though lame from flow :P

IanVS:
Thanks @fbartho.  I'll give @andidev, a bit longer to hopefully test this out, because I have a feeling there could be other issues with Flow, and it would be good to confirm whether that's the case or not.

andidev:
We have a large code base and it seems to work well except this setting. But maybe there is more specific cases related to this setting. But to be honest ai don't think so.

PR #59, opened 2023-04-21
[Breaking] Update dependencies, jest -> vitest, build for Node 16+

IanVS:
I'm preparing to address https://github.com/IanVS/prettier-plugin-sort-imports/issues/22, and as the first step I've created a `next` branch for us to begin working on which will become version 4.0.

This PR also lays some groundwork by updating some dependencies and modernizing a few things:

- Typescript 4.8.4 -> 5.0.4
- Built files ES5 -> ES2021 (Node 16+)
- Jest -> Vitest (tests run in 5 seconds vs 14)
- `.npmignore` -> `files` in package.json (safer)
- Some other minor dep updates

Node 14 is going to be EOL at the end of this month, and I don't think there's a reason to keep shipping ES5, which is very old by now, so I set up the tsconfig to support Node 16.  That's actually the only real user-facing change in this PR, and I'm willing to reconsider it if anyone feels differently about it.


fbartho (approved):
This looks great! I was briefly surprised by the quantity of changes, but then realized that those were all consequences of switching from jest to Vitetest. I have no experience with vitetest, but it seems like developer experience barely changes.

I'm motivated by #22 so if you're ready to move forwards on that one, I'd be happy to do some of it or at least support you if there aren't obvious ways to split up the work.

If I didn't share clearly before, I had done a PR to propose inserting this plugin into my company's monorepo when I first got involved, and then backed it off because I was blocked by the incremental features and bug fixes that were absolute blockers for our repo. Then we had #22 proposed, and I was going to wait until that was done before rolling it out. Then life delayed me from pushing on #22!

For what it's worth I had either unanimous approval, or neutral responses from my team, so I'm definitely motivated to get this included again.

IanVS:
Thanks for the quick review!

So you mean to tell me you've been spending all that work helping to maintain this package and you're not even using it?  😅  I appreciate the help all the more.

My plan for now is to work on removing one option at a time, because I think having multiple PRs open will quickly lead to merge conflicts.  I'd also like to start releasing alpha versions to get early feedback before the entire set of breaking changes is finished.  I think the most helpful thing for now will be reviews, but maybe after a few I'll take a break and you can jump in and work on a few changes, if you have time.  I don't think the removals will be too hard, it will likely be tests that are the most annoying to update.  But we'll see how it goes!

IanVS:
(I'll also take another look at https://github.com/IanVS/prettier-plugin-sort-imports/pull/28 and maybe we can get that one merged in as well)

fbartho:
> So you mean to tell me you've been spending all that work helping to maintain this package and you're not even using it? 😅 I appreciate the help all the more.

Yeah! I definitely would prefer to maintain/contribute to libraries that I use, haha. Let’s get the breaking release rolled together!

Once the majority of changes are ready, I can apply it to our repo to see if we discover any issues.

fbartho:
I have pushed an update to #28 that retargets it at the `next` branch, and also merges your current `next`-status!

PR #62, opened 2023-04-23
[Breaking] Remove importOrderSeparation option

IanVS:
This PR removes the `importOrderSeparation` option.  It can be re-created by adding custom separator groups in the `importOrder` array.  I needed to add a little bit of logic to the code to support this, mostly around handling side-effect and prettier-ignore groups.

To review, I recommend looking at this one commit-by-commit.  I showed in the first commit that our snapshots did not change when updating the `importOrder` array to add separators.  Then, to clean up the tests, I removed those custom separators, and removed some tests and added others.  

Note, we don't have perfect control here, since we don't have a `<BUILTIN_MODULES>` magic word, but I think we can get along without it, since adding a `""` at the start of the `importOrder` array will separate them from the rest (see the `prettier.config.js` in this repo and the effect on `constants.ts`).


fbartho (approved):
I reviewed it commit-wise as you suggested. Looks good to me! 👍

PR #63, opened 2023-04-23
[Breaking] Remove importOrderCaseInsensitive option

IanVS:
Reference #22 

This removes the `importOrderCaseInsensitive` option, making sorts always case-insensitive.


IanVS:
@blutorange I see you might have moved on to using Rome instead of prettier (https://github.com/diffplug/spotless/pull/1663), would you like me to stop tagging you on these PRs?

fbartho (approved):

PR #65, opened 2023-04-24
[Breaking] Remove importOrderGroupNamespaceSpecifiers option

IanVS:
Ref #22 

This removes the `importOrderGroupNamespaceSpecifiers`, and always sort the specifiers (i.e. the option is always turned on).  

I removed a few tests as well that are no longer useful due to removed options.


fbartho (approved):
Looking good!

PR #67, opened 2023-04-25
Replace `importOrderCombineTypeAndValueImports` with `importOrderTypeScriptVersion`

IanVS:
Ref https://github.com/IanVS/prettier-plugin-sort-imports/issues/22

This removes the `importOrderCombineTypeAndValueImports` option, and defaults merging to true for Flow projects (technically anything other than TypeScript, but Flow is the only one).  

For TypeScript, we need to be sure that we can use the `import {type Foo} from …` syntax, which was only added in 4.5.  To that end, this adds a new option, `importOrderTypeScriptVersion`, which the user should set to the version of TS they're using in their project, and we will use that to generate appropriate import statements.  And possibly in the future we can make other adjustments based on the TS version as well.  Detecting the version automatically is tricky, but maybe we can try doing that sometime in the future as well. 

By default, we set the TS version to something ridiculously low, `1.0.0`, meaning that by default we won't combine type and value unless the user sets their version.


README.md
```diff
@@ -129,12 +129,10 @@ module.exports = {
     semi: true,
     importOrder: ['^@core/(.*)$', '', '^@server/(.*)$', '', '^@ui/(.*)$', '', '^[./]'],
     importOrderParserPlugins: ['typescript', 'jsx', 'decorators-legacy'],
-    importOrderCombineTypeAndValueImports: true,
+    importOrderCombineTypeAndValueImports: false, // Disable if using TS 4.4 or below
```
fbartho, line 132: Should this be deleted?

src/index.ts
```diff
@@ -37,12 +37,12 @@ export const options: Record<
         default: [{ value: ['typescript', 'jsx'] }],
         description: 'Provide a list of plugins for special syntax',
     },
-    importOrderCombineTypeAndValueImports: {
-        type: 'boolean',
+    importOrderTypeScriptVersion: {
+        type: 'string',
         category: 'Global',
-        default: false,
+        default: '1.0.0',
```
fbartho, line 43: *Should* we default to 1.0.0?

This is a breaking change, so we could default to 5.0.0 instead

fbartho (approved):
No technical objections! Just a vote to default to 4.5 — any users that already use this plugin have to change their prettier confit from their old flag to this one anyhow, and any new users that are using old typescript can pay the first time config tax.

src/index.ts
fbartho: https://github.com/IanVS/prettier-plugin-sort-imports/issues/22#issuecomment-1522049180

We could alternately not default unless we can't detect :P

IanVS: I responded.  I think that there are risks we'll detect the wrong version, but I think I'm okay with that as long as we keep this option as an override / escape hatch in case the auto-detection is incorrect.  

What do you think about running with just the explicit version for right now, and adding in auto-detection in another PR?

fbartho: No objections.

I just wanted to advocate for the thing that helps people that are starting new projects! That's likely the first spike of interest in a tool like this, followed by new people joining mature (or stagnant) complex codebases.

I think that if people that are stuck on old versions, that's not a surprise to them, and they will notice the "1 other option" that tells them to pass their TypeScript version if they have a complex setup!

I also think it's unlikely that typescript will drastically change import syntax in the future -- it has to stay in sync with the wider JS Community there.

Anyhow, feel free to proceed without auto-detection now. Technically that can always be added in a future release. Technically a breaking behavior change, but… I wouldn't follow SemVer when changing that -- just my 2 cents.

PR #71, opened 2023-05-11
Fix behavior around comments embedded in imports

### LINKED ISSUE #9: Comments on same line as import are moved down
**Your Environment**

-   **Prettier version**: 2.5.1
-   **node version**: 16.13.1

**Describe the bug**

We have the same issue as https://github.com/trivago/prettier-plugin-sort-imports/issues/149, on the latest version 3.3.1.

**To Reproduce**

See issue mentioned above.

### LINKED ISSUE #54: Line break removed when importOrderCombineTypeAndValueImports is used
**Your Environment**

-   **Prettier version**: 2.7.1
-   **node version**: v16.17.0
-   **package manager**: npm@9.3.1
-   **IDE**: Webstorm

**Describe the bug**
Sorting imports with option `importOrderCombineTypeAndValueImports: true` removes a line break at top of prettier anotation if the type import is defined above import line it is merged with.

image

becomes:
image

**Expected behavior**
Expected a line break like this:
image

```
/**
 * @prettier
 */
module.exports = {
    printWidth: 200,
    tabWidth: 4,
    useTabs: false,
    semi: true,
    singleQuote: true,
    jsxSingleQuote: true,
    trailingComma: 'es5',
    arrowParens: 'avoid',
    requirePragma: true,
    bracketSpacing: true,
    quoteProps: 'consistent',
    parser: 'typescript',
    overrides: [
        {
            files: 'src/i18n/*.js',
            options: {
                printWidth: 20,
            },
        },
        {
            files: 'src/i18n/translate.js',
            options: {
                printWidth: 200,
            },
        },
    ],
    importOrder: ['<THIRD_PARTY_MODULES>', '^rootpath$', '^(scripts/.*)$', '(^src/.*)$', '^(test/.*)$'],
    importOrderSeparation: false,
    importOrderSortSpecifiers: true,
    importOrderBuiltinModulesToTop: true,
    importOrderMergeDuplicateImports: true,
    importOrderCombineTypeAndValueImports: true,
};

```

-   [ ] I'm willing to fix this bug 🥇

fbartho:
Fixes #9
Fixes #54

While attempting to work on #9 I found a lot of gnarly edge cases \o/.

Turns out our previous comment-adjuster didn't work for most cases (indeed it thought it was mutating nodes when it was actually generating clean nodes that it was then throwing away 🤦🏻‍♂️, and what "correct" behavior we were seeing before was incredibly brittle and lucky).

I've now added a section to the README that talks about how it handles comments, but generally speaking here's a quick tl;dr: 
1. Comments at top of File and below last import (> 1 line separation) are now preserved in those positions instead of being moved around when the nearest import is re-sorted/merged/moved.
1. Single-line-comments on `ImportDeclaration` or some flavor of `ImportSpecifier` are properly sorted with their parent-node now!
1. "loose" comments generally attach to the next-subsequent `ImportDeclaration` or `ImportSpecifier`. This means that we can now have a leading comment directly before an import, and it will follow that import. **But** it also meant that a ton of our test-cases had "top-of-file" comments that were interpreted as being attached to the first import, so I patched those test-cases so there's a new-line.

I think the new rules are internally consistent, but I await your feedback!

Please help me with the wording in the README and/or feel free to rewrite it, or use the bullets here to improve the phrasing. I thought about hot-linking to exact examples (snapshots?), but I wasn't sure if that was a good idea. We could use &lt;details&gt; components to embed them directly in the readme?

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,6 +34,7 @@ Since then more critical features & fixes have been added, and the options have 
     - [`importOrderTypeScriptVersion`](#importordertypescriptversion)
     - [`importOrderParserPlugins`](#importorderparserplugins)
   - [Prevent imports from being sorted](#prevent-imports-from-being-sorted)
+  - [Comments](#comments)
 - [FAQ / Troubleshooting](#faq--troubleshooting)
 - [Compatibility](#compatibility)
 - [Contribution](#contribution)
@@ -271,16 +272,14 @@ This will keep the `zealand` import at the top instead of moving it below the `a
 entire import statements can be ignored, line comments (`// prettier-ignore`) are recommended over inline comments
 (`/* prettier-ignore */`).
 
-### Commented Lines inside of Imports
+### Comments
 
-We make the following attempts at keeping comments in your imports clean, but there's some pain points when imports are shuffled.
+We make the following attempts at keeping comments in your imports clean:
 
-Specific cases we handle:
-
-- If you have one or more comments that end above the line immediately preceding your first import, we will treat them as top-of-file comment(s) and avoid moving them when the first-import moves down.
-- If you have comments that come after after your last import; those comments and following code will stay below the imports. (Runtime-code between imports will be moved below all the imports)
-- In general, if you place a single-line comment on the same line as an Import `Declaration` or `*Specifier`, we will keep it attached to that same specifier if that line moves around (due to mergers, or sorting changes due to newly inserted imports).
-- Other comments are preserved, and are generally considered `leadingComments` for the subsequent Import `Declaration` or `*Specifier`
+- If you have one or more comments at the top of the file, we will keep them at the top as long as there is a blank line before your first import statement.
+- Comments on lines after the final import statement will not be moved. (Runtime-code between imports will be moved below all the imports).
+- In general, if you place a comment on the same line as an Import `Declaration` or `*Specifier`, we will keep it attached to that same specifier if it moves around.
+- Other comments are preserved, and are generally considered "leading" comments for the subsequent Import `Declaration` or `*Specifier`.
 
 ## FAQ / Troubleshooting
 
--- a/README.md
+++ b/README.md
@@ -271,6 +271,18 @@ This will keep the `zealand` import at the top instead of moving it below the `a
 entire import statements can be ignored, line comments (`// prettier-ignore`) are recommended over inline comments
 (`/* prettier-ignore */`).
 
+### Commented Lines inside of Imports
+
+We make the following attempts at keeping comments in your imports clean, but there's some pain points when imports are shuffled.
+
+Specific cases we handle:
+
+- If you leave a gap after a comment at the top of your file, we will avoid moving it around if the imports below it shift.
+- If you have comments that come after after your last import (with a gap); that comment and following code will stay below the imports.
+- In general, if you place a single-line `CommentLine` (eg. `// …`) on the same line as an `ImportDeclaration` or `ImportSpecifier`, we will keep it attached to that same specifier if that line moves around (due to mergers, or sorting changes due to newly inserted imports).
+- Multi-line-capable `CommentBlocks` (eg. `/** … */`) are automatically formatted to be on a new line by Prettier, so we do not change this behavior.
+- Other comments are preserved, and are generally considered `leadingComments` for the subsequent `ImportDeclaration` or `ImportSpecifier`
+
 ## FAQ / Troubleshooting
 
 Having some trouble or an issue? You can check [FAQ / Troubleshooting section](./docs/TROUBLESHOOTING.md).
--- a/README.md
+++ b/README.md
@@ -279,9 +279,8 @@ Specific cases we handle:
 
 - If you leave a gap after a comment at the top of your file, we will avoid moving it around if the imports below it shift.
 - If you have comments that come after after your last import (with a gap); that comment and following code will stay below the imports.
-- In general, if you place a single-line `CommentLine` (eg. `// …`) on the same line as an `ImportDeclaration` or `ImportSpecifier`, we will keep it attached to that same specifier if that line moves around (due to mergers, or sorting changes due to newly inserted imports).
-- Multi-line-capable `CommentBlocks` (eg. `/** … */`) are automatically formatted to be on a new line by Prettier, so we do not change this behavior.
-- Other comments are preserved, and are generally considered `leadingComments` for the subsequent `ImportDeclaration` or `ImportSpecifier`
+- In general, if you place a single-line comment on the same line as an Import `Declaration` or `*Specifier`, we will keep it attached to that same specifier if that line moves around (due to mergers, or sorting changes due to newly inserted imports).
+- Other comments are preserved, and are generally considered `leadingComments` for the subsequent Import `Declaration` or `*Specifier`
 
 ## FAQ / Troubleshooting
 
--- a/README.md
+++ b/README.md
@@ -277,8 +277,8 @@ We make the following attempts at keeping comments in your imports clean, but th
 
 Specific cases we handle:
 
-- If you leave a gap after a comment at the top of your file, we will avoid moving it around if the imports below it shift.
-- If you have comments that come after after your last import (with a gap); that comment and following code will stay below the imports.
+- If you have one or more comments that end above the line immediately preceding your first import, we will treat them as top-of-file comment(s) and avoid moving them when the first-import moves down.
+- If you have comments that come after after your last import; those comments and following code will stay below the imports. (Runtime-code between imports will be moved below all the imports)
 - In general, if you place a single-line comment on the same line as an Import `Declaration` or `*Specifier`, we will keep it attached to that same specifier if that line moves around (due to mergers, or sorting changes due to newly inserted imports).
 - Other comments are preserved, and are generally considered `leadingComments` for the subsequent Import `Declaration` or `*Specifier`
 
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -37,3 +37,23 @@ const PRETTIER_PLUGIN_SORT_IMPORTS_NEW_LINE =
 export const newLineNode = expressionStatement(
     stringLiteral(PRETTIER_PLUGIN_SORT_IMPORTS_NEW_LINE),
 );
+export const injectNewlinesRegex = /"PRETTIER_PLUGIN_SORT_IMPORTS_NEW_LINE";/gi;
+
+/**
+ * If you have a trailing single-line comment on an ImportSpecifier, and attached to
+ *  the following ImportSpecifier you have a leading single-line comment,
+ * Then Babel's code-generator might render the two comments on the same line.
+ *
+ * This probably should be reported upstream, but for now we have a workaround.
+ */
+export const PATCH_BABEL_GENERATOR_DOUBLE_COMMENTS_ON_ONE_LINE_ISSUE = true;
+
+export const forceANewlineForImportSpecifiersWithLeadingComments = () => ({
+    type: 'CommentLine' as const,
+    value: 'PRETTIER_PLUGIN_SORT_IMPORTS_SPECIFIER_LEADING_COMMENT_PATCH',
+    start: -1,
+    end: -1,
+    loc: { start: { line: -1, column: -1 }, end: { line: -1, column: -1 } },
+});
+export const forceANewlineForImportSpecifiersWithLeadingCommentsRegex =
+    /\/\/PRETTIER_PLUGIN_SORT_IMPORTS_SPECIFIER_LEADING_COMMENT_PATCH/gi;
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -34,29 +34,18 @@ export const TYPES_SPECIAL_WORD = '<TYPES>';
 const PRETTIER_PLUGIN_SORT_IMPORTS_NEW_LINE =
     'PRETTIER_PLUGIN_SORT_IMPORTS_NEW_LINE';
 
+/** Use this to force a newline at top-level scope (good for newlines generated between import blocks) */
 export const newLineNode = expressionStatement(
     stringLiteral(PRETTIER_PLUGIN_SORT_IMPORTS_NEW_LINE),
 );
-export const injectNewlinesRegex = /"PRETTIER_PLUGIN_SORT_IMPORTS_NEW_LINE";/gi;
-
-/**
- * If you have a trailing single-line comment on an ImportSpecifier, and attached to
- *  the following ImportSpecifier you have a leading single-line comment,
- * Then Babel's code-generator might render the two comments on the same line.
- *
- * Additionally, if you have a single-line-comment trailing an ImportDeclaration,
- *  babel might pull
- *
- * This probably should be reported upstream, but for now we have a workaround.
- */
-export const PATCH_BABEL_GENERATOR_DOUBLE_COMMENTS_ON_ONE_LINE_ISSUE = true;
-
-export const forceANewlineForImportsWithAttachedSingleLineComments = () => ({
+/** Use this if you want to force a newline, but you're attaching to leading/inner/trailing Comments */
+export const forceANewlineUsingACommentStatement = () => ({
     type: 'CommentLine' as const,
-    value: 'PRETTIER_PLUGIN_SORT_IMPORTS_SINGLE_LINE_COMMENTS_PATCH',
+    value: 'PRETTIER_PLUGIN_SORT_IMPORTS_NEWLINE_COMMENT',
     start: -1,
     end: -1,
     loc: { start: { line: -1, column: -1 }, end: { line: -1, column: -1 } },
 });
-export const forceANewlineForImportsWithAttachedSingleLineCommentsRegex =
-    /\/\/PRETTIER_PLUGIN_SORT_IMPORTS_SINGLE_LINE_COMMENTS_PATCH/gi;
+
+export const injectNewlinesRegex =
+    /("PRETTIER_PLUGIN_SORT_IMPORTS_NEW_LINE";|\/\/PRETTIER_PLUGIN_SORT_IMPORTS_NEWLINE_COMMENT)/gi;
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -44,16 +44,19 @@ export const injectNewlinesRegex = /"PRETTIER_PLUGIN_SORT_IMPORTS_NEW_LINE";/gi;
  *  the following ImportSpecifier you have a leading single-line comment,
  * Then Babel's code-generator might render the two comments on the same line.
  *
+ * Additionally, if you have a single-line-comment trailing an ImportDeclaration,
+ *  babel might pull
+ *
  * This probably should be reported upstream, but for now we have a workaround.
  */
 export const PATCH_BABEL_GENERATOR_DOUBLE_COMMENTS_ON_ONE_LINE_ISSUE = true;
 
-export const forceANewlineForImportSpecifiersWithLeadingComments = () => ({
+export const forceANewlineForImportsWithAttachedSingleLineComments = () => ({
     type: 'CommentLine' as const,
-    value: 'PRETTIER_PLUGIN_SORT_IMPORTS_SPECIFIER_LEADING_COMMENT_PATCH',
+    value: 'PRETTIER_PLUGIN_SORT_IMPORTS_SINGLE_LINE_COMMENTS_PATCH',
     start: -1,
     end: -1,
     loc: { start: { line: -1, column: -1 }, end: { line: -1, column: -1 } },
 });
-export const forceANewlineForImportSpecifiersWithLeadingCommentsRegex =
-    /\/\/PRETTIER_PLUGIN_SORT_IMPORTS_SPECIFIER_LEADING_COMMENT_PATCH/gi;
+export const forceANewlineForImportsWithAttachedSingleLineCommentsRegex =
+    /\/\/PRETTIER_PLUGIN_SORT_IMPORTS_SINGLE_LINE_COMMENTS_PATCH/gi;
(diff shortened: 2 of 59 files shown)


fbartho:
@IanVS Happy to report this is green in CI! Now just needs to go ✅ in our hearts.

src/constants.ts
```diff
@@ -37,3 +37,26 @@ const PRETTIER_PLUGIN_SORT_IMPORTS_NEW_LINE =
...
 );
+export const injectNewlinesRegex = /"PRETTIER_PLUGIN_SORT_IMPORTS_NEW_LINE";/gi;
+
+/**
+ * If you have a trailing single-line comment on an ImportSpecifier, and attached to
+ *  the following ImportSpecifier you have a leading single-line comment,
+ * Then Babel's code-generator might render the two comments on the same line.
+ *
+ * Additionally, if you have a single-line-comment trailing an ImportDeclaration,
+ *  babel might pull
```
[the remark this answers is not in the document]
IanVS: Actually, I was able to find https://github.com/babel/babel/pull/14979, which was included in a more recent babel version that we were using.  

I know you put a bunch of thought into this, so I kind of hate to say it, but I updated babel in your branch (and fixed a few minor typescript issues), and this `PATCH_BABEL_GENERATOR_DOUBLE_COMMENTS_ON_ONE_LINE_ISSUE` no longer seems to be needed.  

In fact, it seems like now is a good time to update babel as part of the breaking change, since it may impact the way we do formatting.  I'll push that up as a separate PR, so we can examine any other impacts to existing snapshots.  Then we can apply this branch on top of it, and maybe some of the hoops you had to jump through to get comments working correctly won't be needed anymore.

Sorry I didn't think to upgrade babel sooner.  Because we use a caret in the range, new installations will already be using the new versions.  I start to understand why the upstream project locked their version of babel, but I still think  it's not the best thing to do, and causes more issues than it potentially solves.

IanVS (commented):
Really great work here!  I haven't reviewed the whole thing yet, because I think there's a chance we might be able to simplify some of it based on improvements made to babel itself.

src/constants.ts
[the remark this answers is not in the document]
IanVS: FWIW, it looks like ASTExplorer is using an older version of babel too.  🤦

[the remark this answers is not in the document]
fbartho: Yeah, new babel fixes some of the root issues! Removed my workarounds!

README.md
```diff
@@ -271,6 +271,18 @@ This will keep the `zealand` import at the top instead of moving it below the `a
...
+### Commented Lines inside of Imports
+
+We make the following attempts at keeping comments in your imports clean, but there's some pain points when imports are shuffled.
+
+Specific cases we handle:
+
+- If you leave a gap after a comment at the top of your file, we will avoid moving it around if the imports below it shift.
+- If you have comments that come after after your last import (with a gap); that comment and following code will stay below the imports.
+- In general, if you place a single-line `CommentLine` (eg. `// …`) on the same line as an `ImportDeclaration` or `ImportSpecifier`, we will keep it attached to that same specifier if that line moves around (due to mergers, or sorting changes due to newly inserted imports).
+- Multi-line-capable `CommentBlocks` (eg. `/** … */`) are automatically formatted to be on a new line by Prettier, so we do not change this behavior.
```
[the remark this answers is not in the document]
fbartho: It was babel's code-generator, I tested it in isolation -- the right set of nodes would replicate the complaints I had even without our custom sorting code in play. Indeed one of the several issues linked to from this PR: https://github.com/babel/babel/pull/14979 describes the same problems I was running into.

fbartho:
@IanVS Green again with new babel!

fbartho:
> I think there's a chance we might be able to simplify some of it based on improvements made to babel itself.

While I was able to remove my patch/hack/workaround, I wasn't able to fundamentally simplify it because babel doesn't concern itself with "moving nodes". So we still had the problem that comments were associated with multiple Import statements (trailing one, and leading another), and the whole point of our plugin is to move the leaders/trailer nodes around. 

Babel would then naively render the comment the first time it sees it, which would invariably, often lead to unexpected comment positioning (from the user-perspective).

So we have to detach all the comments, and reattach them to *only* the right attachment points.

src/utils/__tests__/adjust-comments-on-sorted-nodes.spec.ts
```diff
@@ -75,16 +75,16 @@ test('it does not move comments at before all import declarations', () => {
...
-    expect(adjustedNodes).toHaveLength(3);
-    expect(leadingComments(adjustedNodes[0])).toEqual([
-        ' comment c1',
-        ' comment c2',
-    ]);
-    expect(trailingComments(adjustedNodes[0])).toEqual([]);
-    expect(leadingComments(adjustedNodes[1])).toEqual([]);
-    expect(trailingComments(adjustedNodes[1])).toEqual([]);
+    expect(adjustedNodes).toHaveLength(4);
+    // Comment c1 is explicitly detached so it stays with the top-of-file
```
IanVS, line 79: Why doesn't comment c1 move?  In your note in the readme, it says"

>If you leave a gap after a comment at the top of your file, we will avoid moving it around if the imports below it shift.

So I would assume since there's no gap, that c1 and c2 would both move with the import from "c".

src/utils/__tests__/adjust-comments-on-sorted-nodes.spec.ts
```diff
@@ -98,11 +98,16 @@ test('it does not affect comments after all import declarations', () => {
...
     const finalNodes = [importNodes[2], importNodes[1], importNodes[0]];
     const adjustedNodes = adjustCommentsOnSortedNodes(importNodes, finalNodes);
-    expect(adjustedNodes).toHaveLength(3);
-    expect(leadingComments(adjustedNodes[0])).toEqual([]);
-    expect(trailingComments(adjustedNodes[0])).toEqual([]);
+    expect(adjustedNodes).toHaveLength(4);
     expect(leadingComments(adjustedNodes[1])).toEqual([]);
-    expect(trailingComments(adjustedNodes[1])).toEqual([]);
+    // "final 1" is attached as a trailing-comment for import from "a"
+    // but "final 2" is detached so it stays with the bottom-of-imports
```
IanVS, line 104: This behavior seems a bit odd to me.  The test description is that comments after all the imports should not be moved, and I think I agree with that. The only kind of "trailing" comment that feels like it should move with an import is one that is on the same line.  Comments that start on the line below an import feel like they shouldn't be associated with that import.

src/utils/__tests__/get-code-from-ast.spec.ts
```diff
@@ -26,10 +24,7 @@ import a from 'a';
         directives: [],
     });
     expect(format(formatted, { parser: 'babel' })).toEqual(
-        `// first comment
-// second comment
-
-import a from "a";
+        `import a from "a";
```
IanVS, line 27: I'm guessing you removed the comments here because they're tested elsewhere and not directly related to this function, right?

tests/Flow/__snapshots__/ppsi.spec.ts.snap
```diff
@@ -33,7 +34,6 @@ export function givesAFoo3Obj(): AliasFoo3 {
 /**
  * @flow
  */
-
```
IanVS, line 36: Ideally we wouldn't remove this newline.  I wonder if we can somehow use the `loc` of all comments at the top of a file that aren't attached to an import.

tests/Flow/__snapshots__/ppsi.spec.ts.snap
```diff
@@ -196,6 +196,8 @@ import {fooValue} from './foo';
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 /**
  * @flow
- */ import { fooValue, type Foo } from "./foo";
+ */
+
+import { fooValue, type Foo } from "./foo";
```
IanVS, line 201: Nice, this is much better!

tests/ImportCommentsPreserved/__snapshots__/ppsi.spec.ts.snap
```diff
@@ -0,0 +1,178 @@
...
+ * This is part of multiple-lines at document-top
+ * Second line of text, comment is followed by a gap (gap gets eaten)
+ */
+
+import b from "b";
+import a from "a";
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+/**
+ * This is part of multiple-lines at document-top
+ * Second line of text, comment is followed by a gap (gap gets eaten)
```
IanVS, line 43: What does it mean, "get gets eaten"?  It looks like it's being preserved here (which feels correct).

tests/ImportCommentsPreserved/__snapshots__/ppsi.spec.ts.snap
```diff
@@ -0,0 +1,178 @@
...
+ */
+
+import a from "a";
+import b from "b";
+
```
IanVS, line 48: Any idea why there's a newline here?

tests/ImportCommentsPreserved/import-comments-document-top-test3.ts
```diff
@@ -0,0 +1,7 @@
+/**
+ * This comment is before the whole block,
+ * it is not attached to the following import.
+ */
+
+import b from "b";
+import a from "a";
```
IanVS, line 7: How is this one different from `import-comments-document-top-test2?`

tests/ImportCommentsPreserved/__snapshots__/ppsi.spec.ts.snap
```diff
@@ -0,0 +1,178 @@
...
+`;
+
+exports[`import-comments-outer-comments-block-comments.ts - typescript-verify > import-comments-outer-comments-block-comments.ts 1`] = `
+// Loose leading comment before imports (should not be dragged down with B)
+
+/* Leading comment before B */
+import B from "b"; /* Trailing comment on same-line as B (babel doesn't like CommentBlocks on same-line as anything else, so it will be shifted to a clean line) */
```
IanVS, line 74: Hmmm, it would be great to keep this together with the import.  Where its ending up feels a little broken.

tests/ImportCommentsPreserved/__snapshots__/ppsi.spec.ts.snap
```diff
@@ -0,0 +1,178 @@
...
+
+// Loose trailing comment after imports (should not be dragged up with A)
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+// Loose leading comment before imports (should not be dragged down with B)
+
+/* Trailing comment on same-line as B (babel doesn't like CommentBlocks on same-line as anything else, so it will be shifted to a clean line) */
+/* Trailing comment on first line after B (This is treated as a leading comment for next import A!) */
+
+/* Leading comment before A */
+import A from "a"; /* Trailing comment on same-line as A (babel doesn't like CommentBlocks on same-line as anything else, so it will be shifted to a clean line) */
```
IanVS, line 90: Oddly, this one is _not_ shifted to a new line.

tests/ImportCommentsPreserved/__snapshots__/ppsi.spec.ts.snap
```diff
@@ -0,0 +1,178 @@
...
+import {
+    h1, // h1 same-line single-line-comment
+    h2, // h2 same-line single-line-comment
+    // h3 leading single-line-comment
+    h3, // h3 same-line single-line-comment
+    h4, // h4 same-line single-line-comment
+} from "h";
+import {
+    i1 /* i1 same-line will be treated as trailing comment for i1 */,
+    i2 /* i2 same-line will be treated as leading comment for i1 */,
```
IanVS, line 164: This comment seems worded incorrectly, it's kept as a same-line trailing comment on i2 (which is great!)

tests/ImportsNotSeparated/__snapshots__/ppsi.spec.ts.snap
```diff
@@ -96,10 +99,11 @@ function add(a:number,b:number) {
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 // I am top level comment in this file.
 // I am second line of top level comment in this file.
+
 import "./commands";
 import React from "react";
-
 // Comment
+
```
IanVS, line 106: 🤨 where did this newline come from?

tests/ImportsNotSeparated/__snapshots__/ppsi.spec.ts.snap
```diff
@@ -164,7 +170,7 @@ import thirdParty from "third-party";
 import {
     random // inner comment
 } from './random';
-// leading comment
+// trailing comment for import-random (export statement is code that will be sorted to below imports, and so doesn't have precedence over the import-comment-attachment rules!)
```
IanVS, line 173: Is this a side-effect of how things are working, or a behavior that you think is correct?  

Personally, it seems to me like the only kind of trailing comments we should keep with an import are the ones that start on the same line as the end of the import declaration.  Anything else feels unexpected, since I never put a comment on a line after something that it's associated with.  But maybe some people do that!

tests/RemoveStatementsFromOriginalCode/__snapshots__/ppsi.spec.ts.snap
```diff
@@ -27,10 +27,8 @@ import { bar } from "a";
 
 function baz() {
 }~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
-/* foo */
-
 import { bar } from "a";
-import { foo } from "c";
+/* foo */ import { foo } from "c";
```
IanVS, line 31: Nice!

IanVS (commented):
I reviewed all the easy stuff (mostly the tests, which are excellent, by the way).  I have some questions, mostly about trailing comments on lines after the import declaration.  I'll review the code itself tomorrow, which will probably give me a better understanding of why things are working the way they are.

src/utils/__tests__/adjust-comments-on-sorted-nodes.spec.ts
fbartho: Sorry, I see how I confused things particularly with my variable names and the term “gap”.

I didn’t mean “gap” as in blank line — I have to consider each comment independently (or I need even more richness in my data model/scan-passes). — So gap is “if this comment and the import-node were the only things that existed, would there be a gap?”

So “leading gap” really is “is this specific 1-line comment starting

src/utils/__tests__/adjust-comments-on-sorted-nodes.spec.ts
fbartho: I could accept that argument.

If you want to change that behavior, it’s just removing the “+ 1” on the `trailingGap` comparison (plus updating Readme and test-cases)

src/utils/__tests__/get-code-from-ast.spec.ts
fbartho: Exactly!

tests/ImportCommentsPreserved/__snapshots__/ppsi.spec.ts.snap
fbartho: This comment was added before today’s changes, gap no longer gets eaten.

I’ll fix the wording!

tests/ImportCommentsPreserved/__snapshots__/ppsi.spec.ts.snap
fbartho: This comment was added before today’s changes, it no-longer gets force-wrapped.

I’ll fix the wording!

tests/ImportCommentsPreserved/__snapshots__/ppsi.spec.ts.snap
fbartho: This comment was added before today’s changes, it no-longer gets force-wrapped.

I’ll fix the wording!

tests/ImportsNotSeparated/__snapshots__/ppsi.spec.ts.snap
fbartho: A newline appears when dealing with a “bottom-of-imports” comment.

Your request to change the logic so that nothing attaches (to top-level-decls) as trailing except for same-line comments address this a bit more cleanly?

tests/ImportsNotSeparated/__snapshots__/ppsi.spec.ts.snap
fbartho: The situation here is that `export` is not considered part of the nodes we’re allowed to modify, but we do have logic to support pulling trailing comments around elsewhere, so the comment is preferentially bound.

I’ll tweak things to disable trailing comments except-on-same-line, and then you can review the result.

src/utils/__tests__/adjust-comments-on-sorted-nodes.spec.ts
fbartho: I'm implementing this change now, and then updating the comments/snapshots to match!

tests/Flow/__snapshots__/ppsi.spec.ts.snap
fbartho: This isn't deleted exactly, it's the comment below that was injected as a "second" top-of-file comment.

tests/ImportCommentsPreserved/__snapshots__/ppsi.spec.ts.snap
fbartho: I think when imports get shuffled, one newline escapes and gets inserted below the imports (excess newlines are trimmed).

I'm not sure on this one. Is it a blocking issue?

tests/ImportCommentsPreserved/import-comments-document-top-test3.ts
fbartho: It is not, I had other test cases that made this test file redundant, and I shall clean it up post-haste.

src/utils/__tests__/adjust-comments-on-sorted-nodes.spec.ts
fbartho: Renamed the variables and comments, in particular because the change from https://github.com/IanVS/prettier-plugin-sort-imports/pull/71#discussion_r1191887394 also further twisted the meaning of one of them.

tests/Flow/__snapshots__/ppsi.spec.ts.snap
fbartho: It's not actually being removed so much as a comment is being shuffled to be directly trailing this comment.

tests/ImportsNotSeparated/__snapshots__/ppsi.spec.ts.snap
fbartho: Please check out this case again (after I push up my latest changes)

fbartho (commented):
Somehow started a review on myself [will clear this]

tests/ImportsNotSeparated/__snapshots__/ppsi.spec.ts.snap
fbartho: This issue is gone with the latest commit! https://github.com/IanVS/prettier-plugin-sort-imports/pull/71/commits/fca03777f595716b1ae64753b9393de52dfbaede

fbartho:
✅ Ready for next review!

src/utils/__tests__/adjust-comments-on-sorted-nodes.spec.ts
fbartho: [Done ✅ ]

src/utils/get-comment-registry.ts
```diff
@@ -0,0 +1,431 @@
...
+}: {
+    /** Original declaration, not the re-sorted output-node! */
+    firstImport: ImportDeclaration;
+    /** Original declaration, not the re-sorted output-node! */
+    lastImport: ImportDeclaration;
+
+    /** Constructed Output Nodes */
+    outputNodes: ImportDeclaration[];
+}) => {
+    if ((outputNodes.length === 0 || !firstImport, !lastImport)) {
```
IanVS: I haven't seen this syntax before in an if statement, and I'm not sure I understand it.  But, from playing around, it seems that only the last expression in the sequence is considered.  So, if I have:

```ts
function test(x, y) {
    if (x, y) {
        console.log('true');
    } else {
        console.log('false');
    }
}
```

Then the only way to get `true` is to pass a truthy `y`, and it doesn't matter what `x` is.  

Can you maybe explain what the goal of this check is?

IanVS (approved):
This is awesome work!  I think you've managed to handle all the tricky comment cases in a way that makes good sense.  I pushed up a couple of commits to add a couple of small test cases, reword the README slightly, and clean up some more of the trailing comment logic.  Feel free to push back on any of it, or merge if you're satisfied.

src/utils/get-comment-registry.ts
fbartho: Brain-💨 — that was just supposed to be `||` instead of javascript’s crazy & unnecessary comma operator `,` (that should be banned — eslint-rule?).

You’re right there’s not a test-case.

The way I wrote the code that calls this function, I construct the `lastImport` from the same array that I construct the `firstImport` — so unless you’re calling this helper in new code, it’s not possible to get one of the imports to be non-null while the other is null.

IanVS: Take a look at my commits, I think we don't need `lastImport` at all anymore, so I simplified it down.

fbartho: I looked at all your commits and have no objections with them! I definitely had stuff in there that was "left-overs" from previous implementation/approaches, or just no longer relevant debugging. Thanks for cleaning all that up!

PR #72, opened 2023-05-11
Upgrade Babel to 7.21

IanVS:
This upgrade pulls in https://github.com/babel/babel/pull/14979, which improves the way that Babel handles comments.  I think that https://github.com/IanVS/prettier-plugin-sort-imports/pull/71 will still be necessary, but this PR might have an impact on it, hopefully resulting in less complexity 🤞.


fbartho (approved):
Absolutely zero objections

PR #73, opened 2023-05-11
Update to latest versions of Github Actions

fbartho:
I noticed a deprecation warning in GitHub Actions run summaries:
```
Node.js 12 actions are deprecated. Please update the following actions to use Node.js 16: actions/checkout@v2, actions/setup-node@v2. For more information see: https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/.
```

Thankfully this is a trivial upgrade, I also cleaned up a few other related things in the workflow files!


jasikpark (approved):
lgtm

PR #74, opened 2023-05-12
Clean-up lodash.clone dep, as it serves no purpose

fbartho:
I believe the usage of `lodash.clone` used to be more prevalent (and necessary). But at this point we only had 1-remaining call-site, and I traced it, and don't see any mutating methods along the way, so I'm relatively confident this isn't necessary.

I also tested `JSON.parse(JSON.stringify(…))` and that passed the tests too, so if we'd prefer to use that instead, I can do so, just let me know!

### DIFF
--- a/src/utils/get-sorted-nodes-by-import-order.ts
+++ b/src/utils/get-sorted-nodes-by-import-order.ts
@@ -1,5 +1,3 @@
-import clone from 'lodash.clone';
-
 import {
     BUILTIN_MODULES,
     newLineNode,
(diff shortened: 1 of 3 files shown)


jasikpark:
ig the other approach would be to use `structuredClone`?

fbartho:
@jasikpark yeah! I haven't used `structuredClone` in production yet, even if I'm vaguely familiar with it.

I don't think we need any flavor of clone, but that's an option too if we still think that's necessary.

src/utils/get-sorted-nodes-by-import-order.ts
```diff
@@ -14,13 +12,15 @@ import { getSortedNodesGroup } from './get-sorted-nodes-group';
...
  * @param options Options to influence the behavior of the sorting algorithm.
  */
-export const getSortedNodesByImportOrder: GetSortedNodes = (nodes, options) => {
+export const getSortedNodesByImportOrder: GetSortedNodes = (
+    originalNodes,
+    options,
+) => {
     let { importOrder } = options;
 
-    const originalNodes = nodes.map(clone);
```
IanVS, line 23: Are we sure we don't need to clone?  It looks like the use of `getSortedNodesGroup` has a `.sort`, which will change them in place.  Though, I guess it's not mutating the nodes themselves, just their order within our `importOrderGroups` array.  So yeah, I think I've convinced myself this is fine.

IanVS (approved):

PR #75, opened 2023-05-12
Upgrade to Prettier 3 in devDependencies

fbartho:
This PR started with #43, and then merge-conflict resolution was necessary.

No big worries about this PR (or your prettier-3 PR), but I did have to fix `run_spec.ts` to get the tests to pass, and I don't know why I had to.  -- Please specifically review this commit. https://github.com/IanVS/prettier-plugin-sort-imports/commit/5832d91a3a7e337a03ce1c0562227fd718a47b6e (there was no `output` variable in scope, so I guessed!)


fbartho:
I discovered we have working code for `alpha.4`, but broken code with all of the following versions:
- `3.0.0-alpha.11`
- `3.0.0-alpha.10`
- `3.0.0-alpha.9-for-vscode`
- `3.0.0-alpha.8-for-vscode`
- `3.0.0-alpha.7-for-vscode`
- `3.0.0-alpha.6`
- `3.0.0-alpha.5`

Specifically the error is:
```
{
  "clone": "Function<clone>",
  "code": "BABEL_PARSER_SYNTAX_ERROR",
  "details": {
    "expected": "from",
  },
  "loc": {
    "column": 12,
    "constructor": "Function<R>",
    "index": 65,
    "line": 3,
  },
  "pos": 65,
  "reasonCode": "UnexpectedToken",
}
```

fbartho:
In alpha.5 a new babel-ts parser (to parse TypeScript v5) was introduced in this commit: https://github.com/prettier/prettier/commit/019ebe52078efb6ba0ed802ea104617744f42e4c

I think that's likely related to our problem.

IanVS:
I guess maybe we might have another use for our typescript version flag after all.

fbartho:
@IanVS since you added explicit support for `babel-ts` in https://github.com/IanVS/prettier-plugin-sort-imports/pull/97 maybe there shouldn't be a problem anymore?

fbartho:
I synched up with the status of `next` and unfortunately we still get the same babel-parser syntax error when were on prettier alpha.11.

I looked at our yarn.lock and I can't find babel-ts in there anywhere, so I think that's an internal plugin vended by babel, in which case upgrading our babel/babel-parser version(s) might be required?

Regardless, using the latest versions doesn't help.

The latest error:
```
  "codeFrame": "  1 | import { ComponentProps, useEffect } from "react";
> 2 | import type React from "react";
    |             ^
[…snipped…]
Caused by: SyntaxError: Unexpected token, expected "from" (2:12)
```
Looks to me like it doesn't understand `import type` expressions? aka weird.

IanVS:
` babel-ts` is one of the parser names in `prettier/parser-babel`: https://github.com/prettier/prettier/blob/7a015745c956ad1d00b603b9f851cab8ed052102/src/language-js/parse/babel.js#L251

IanVS (approved):
I think this is ready.  @fbartho where were you seeing the errors before?  

https://github.com/IanVS/prettier-plugin-sort-imports/pull/75/commits/c5b0d943496b33d85a3105c2651ab1d30e5cf6f6 feels bad, but this seems to just be a change in prettier's output, and nothing that we control.

fbartho:
>I think this is ready. @fbartho where were you seeing the errors before?

Tests were failing, so since we’re green now, they must have reverted it! I think this is ready too

>https://github.com/IanVS/prettier-plugin-sort-imports/commit/c5b0d943496b33d85a3105c2651ab1d30e5cf6f6 feels bad, but this seems to just be a change in prettier's output, and nothing that we control.

I concur. This is a prettier regression. We didn’t cause that. Probably worth reporting upstream though! — Note: this might be something related to typescript versions, idk.

PR #80, opened 2023-05-12
Preserve trailing comments on specifiers

### LINKED ISSUE #79: Loses @ts-directives nested inside an Import sometimes
Not filling out the template, just providing a code-snippet and then trying to fix it.

```ts
import { readFileSync } from "fs";
import * as path from "path";
import { PackageJson } from "type-fest";

import {
	isHCAPackageRegexp,
	HCA_PACKAGE_DEFAULT_PREFIX as HCA_PACKAGE_DEFAULT_PREFIX_,
	// @ts-expect-error
} from "./PackageDefsJS";
```

When formatting this, `// @ts-expect-error` gets deleted! You can replicate this if you add it as a test-case in `tests/ImportCommentsPreserved/`

fbartho:
Fixes #79

Trailing comments on specifiers don't get automatically saved like top-level ones do.  This ensures that such comments are not lost

Before:
```ts
import {
    b2,a2,
    // @ts-expect-error
} from "b";

import {
    b1,
    a1,

    // @ts-expect-error
} from "a";
```
After
```ts
import {
    a1,
    b1,
    // @ts-expect-error
} from "a";
import {
    a2,
    b2,
    // @ts-expect-error
} from "b";
```

### DIFF
--- a/src/utils/get-comment-registry.ts
+++ b/src/utils/get-comment-registry.ts
@@ -51,8 +51,10 @@ const orderedCommentKeysToRegister = [
 export interface CommentEntry {
     owner: ImportDeclaration | SomeSpecifier;
     ownerIsSpecifier: boolean;
-    // Special case for leaving comments at top-of-file
+    /** Special case for leaving comments at top-of-file */
     needsTopOfFileOwner?: boolean;
+    /** Comments that follow the last specifier must stay at the bottom of their import block! */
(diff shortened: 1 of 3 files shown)


fbartho:
I think we could potentially merge with this PR and see if it bugs anyone in the wild? idk, at least we're not losing important comments.

IanVS (approved):

PR #82, opened 2023-05-14
Behavior Change: All comments before firstImport are treated as top-of-file

### LINKED ISSUE #81: Behavior Change: Protect all comments before the first import (assume they're pragmas/top-of-file)
In https://github.com/IanVS/prettier-plugin-sort-imports/issues/54#issuecomment-1546917730 we decided to protect top-of-file (before-first-import) comments more aggressively.

Unhappy Test Case:
```js
/**
 * @prettier
 */
import { b } from 'b';
import type { A } from 'a';
import { a } from 'a';

// Should not become: ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

import { a, type A } from "a";
/**
 * @prettier
 */
import { b } from "b";

```

fbartho:
Fixes #81 

- This also reverts my `alpha.4` behavior-change where top-of-file comments always got a newline injected below them. Now we neither inject nor delete 1 blank line below the last top-of-file-comment.
- This reverts snapshot changes (around top-of-file blank-lines) as compared to our pre-4.0.0-alpha.4 behavior.

### DIFF
--- a/tests/ImportCommentsPreserved/__snapshots__/ppsi.spec.ts.snap
+++ b/tests/ImportCommentsPreserved/__snapshots__/ppsi.spec.ts.snap
@@ -247,6 +247,22 @@ import {
 
     // @ts-expect-error
 } from "a";
+import {
+
+    damn_long_1,
+    damn_long_2,
+
+    damn_long_3,
+    damn_long_4,
+
+
+
+    damn_long_5,
+    damn_long_6,
+    damn_long_7,
+    damn_long_8,
+
+} from "d_proving_all_specifier_blank_lines_get_removed";
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 import {
     a1,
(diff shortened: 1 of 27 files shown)


IanVS (commented):
I pushed some changes to allow us to preserve the spacing between comments at the tops of files (which I think is important).  It got pretty tricky, but I think it's working now for the majority of cases.  

I also had to add some logic to handle type+value import merging.  Previously, we would start with values, and try merging types in.  But if the first import at the top of the page is a type and it gets merged with a value, then we basically get rid of the first import, which was causing problems.  I think the changes I made here should fix the behavior explained at the end of https://github.com/IanVS/prettier-plugin-sort-imports/issues/54.

tests/ImportCommentsPreserved/__snapshots__/ppsi.spec.ts.snap
```diff
@@ -33,16 +33,22 @@ exports[`import-comments-document-top-all-are-untouched.ts - typescript-verify >
...
 import { a } from 'a';
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 /**
  * @prettier
  * This comment is before the first import, and so should never be moved.
  * https://github.com/IanVS/prettier-plugin-sort-imports/issues/81
  */
+
+// second comment
+
```
fbartho: Was this injected comment line intentional?

IanVS (approved):

PR #83, opened 2023-05-15
Update our Default Sort Order builtins, (thirdparty & absolute paths), relative paths

### LINKED ISSUE #76: Improve default `importOrder`
Currently, the default `importOrder` that we ship is kind of worthless, and only sorts everything alphabeteically.

As discussed in https://github.com/IanVS/prettier-plugin-sort-imports/issues/22, let's make the default order be:

```js
importOrder: [
	// builtins (already happens no matter what the `importOrder` is set to)
	'<THIRD_PARTY_MODULES>',
	'^[./]', // Relative (local) imports
]
```

We could get fancier, but I think this is a sane default that most users will benefit from, and they're able to customize it further if they want.

fbartho:
Fixes #76 

```js
[
            // built-ins are implicitly first
    '<THIRD_PARTY_MODULES>', // Actually means "all things that aren't otherwise matched" aka THIRD_PARTY_IMPORTS & Absolute Path Imports
    '^[.]', // relative imports
],
```

- There's no way to separate THIRD_PARTY_MODULES and absolute-imports currently
- We can discuss a renaming for THIRD_PARTY_MODULES later

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -4,6 +4,7 @@ import { parsers as flowParsers } from 'prettier/parser-flow';
 import { parsers as htmlParsers } from 'prettier/parser-html';
 import { parsers as typescriptParsers } from 'prettier/parser-typescript';
 
+import { THIRD_PARTY_MODULES_SPECIAL_WORD } from './constants';
 import { defaultPreprocessor } from './preprocessors/default';
 import { vuePreprocessor } from './preprocessors/vue';
 import type { PrettierOptions } from './types';
@@ -25,8 +26,17 @@ export const options: Record<
         type: 'path',
         category: 'Global',
         array: true,
-        default: [{ value: [] }],
-        description: 'Provide an order to sort imports.',
+        default: [
+            {
+                value: [
+                    // built-ins are implicitly first
+                    THIRD_PARTY_MODULES_SPECIAL_WORD,
+                    '^[.]', // relative imports
+                ],
+            },
+        ],
+        description:
+            'Provide an order to sort imports. [built-ins are always implicitly first]',
     },
     importOrderParserPlugins: {
         type: 'path',
--- a/src/index.ts
+++ b/src/index.ts
@@ -29,14 +29,14 @@ export const options: Record<
         default: [
             {
                 value: [
-                    // built-ins are implicitly first
-                    THIRD_PARTY_MODULES_SPECIAL_WORD,
+                    // node.js built-ins are always first
+                    THIRD_PARTY_MODULES_SPECIAL_WORD, // Everything not matching relative imports
                     '^[.]', // relative imports
                 ],
             },
         ],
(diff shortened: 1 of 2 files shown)


IanVS:
>There's no way to separate THIRD_PARTY_MODULES and absolute-imports currently

When you say "absolute-imports", do you mean imports using an absolute file path, like `/users/foo/module`?  Or, do you mean internal (non-third-party) imports that are aliased and thus are non-relative like `@src/helpers/module` (ref https://github.com/trivago/prettier-plugin-sort-imports/issues/162)?

fbartho:
I mean “libs/path/to/thing.ts” vs (`../../to/thing.ts` or ‘./libs/path/to/thing.ts”)

fbartho:
I guess I should have called that a “root-relative-path” or repo-root-absolute-path? Idk what the right name for it is, but it is a pattern (that I discouraged in my repo(s)), that people ask for often enough that there’s instructions on how to use them in various runtime, typescript, webpack, etc.

fbartho:
> Or, do you mean internal (non-third-party) imports that are aliased and thus are non-relative like @src/helpers/module (ref https://github.com/trivago/prettier-plugin-sort-imports/issues/162)?

Yes, this. It’s clear it’s bedtime because I totally missed your comment about this.

IanVS:
I pushed up some examples for the README, let me know what you think, and feel free to adjust them.

fbartho:
No objection to your README changes!

src/index.ts
```diff
@@ -25,8 +26,17 @@ export const options: Record<
...
         category: 'Global',
         array: true,
-        default: [{ value: [] }],
-        description: 'Provide an order to sort imports.',
+        default: [
+            {
+                value: [
+                    // node.js built-ins are always first
+                    THIRD_PARTY_MODULES_SPECIAL_WORD, // Everything not matching relative imports
+                    '^[.]', // relative imports
```
IanVS, line 34: Should this be `^[./]`?  That seems to be what we use in our tests, at least...

fbartho: I don’t think so?

- `^[./]` matches `/foo/bar`, `../foo/bar` , `./foo/bar` 
- `^[.] only matches `../foo/bar`, `./foo/bar` — and doesn’t match file-system absolute-path imports

This is exactly the confusion I was a little worried about.

fbartho: I don’t think file-system absolute-path imports are that common, but yeah

fbartho: [I see now that you edited your github comment after I had loaded the page, but before I replied, lol]

IanVS: Alright, let's leave it the way you've got it.  Absolute file system imports would be weird, and should probably deserve an explicit importOrder.

IanVS (approved):

PR #84, opened 2023-05-15
Slight reorganization around top-of-file comments

fbartho:
I think this makes the code a little easier to follow.

I wasn't able to add an end-to-end integration test around "empty-statements-at-top-of-file" because you're right @IanVS, other layers prevent that from happening.

Instead I added a unit-test to assert the fatal-errors that we throw.

### DIFF
--- a/src/utils/get-comment-registry.ts
+++ b/src/utils/get-comment-registry.ts
@@ -244,7 +244,7 @@ export const getCommentRegistryFromImportDeclarations = ({
     /** Constructed Output Nodes */
     outputNodes: ImportDeclaration[];
 }) => {
-    if (outputNodes.length === 0 || !firstImport) {
+    if (outputNodes?.length === 0 || !firstImport) {
         return [];
     }
 
@@ -369,7 +369,12 @@ export function attachCommentsToOutputNodes(
     /** Original declaration, not the re-sorted output-node! */
     firstImport: ImportDeclaration,
 ) {
-    if (outputNodes.length === 0) {
+    if (!Array.isArray(commentEntriesFromRegistry)) {
+        throw new Error(
+            'Fatal Internal Error: Expected a list of commentEntriesFromRegistry',
+        );
+    }
+    if (outputNodes == null || outputNodes.length === 0) {
         // attachCommentsToOutputNodes implies that there's at least one output node so this shouldn't happen
         throw new Error(
             "Fatal Internal Error: Can't attach comments to empty output",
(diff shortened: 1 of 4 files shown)


src/utils/get-comment-registry.ts
```diff
@@ -243,7 +244,7 @@ export const getCommentRegistryFromImportDeclarations = ({
     /** Constructed Output Nodes */
     outputNodes: ImportDeclaration[];
 }) => {
-    if (outputNodes.length === 0 || !firstImport) {
+    if (outputNodes?.length === 0 || !firstImport) {
```
IanVS, line 247: When would `outputNodes` be undefined?  We have TS to guarantee it's an array just above, right?

src/utils/get-comment-registry.ts
```diff
@@ -368,7 +369,12 @@ export function attachCommentsToOutputNodes(
     /** Original declaration, not the re-sorted output-node! */
     firstImport: ImportDeclaration,
 ) {
-    if (outputNodes.length === 0) {
+    if (!Array.isArray(commentEntriesFromRegistry)) {
```
IanVS, line 372: How would this happen without throwing a TypeScript compilation error?

src/utils/get-comment-registry.ts
```diff
@@ -368,7 +369,12 @@ export function attachCommentsToOutputNodes(
...
+        throw new Error(
+            'Fatal Internal Error: Expected a list of commentEntriesFromRegistry',
+        );
+    }
+    if (outputNodes == null || outputNodes.length === 0) {
```
IanVS, line 377: Why check for a null `outputNodes`?

src/utils/get-comment-registry.ts
fbartho: I concur that TypeScript would prevent that. 

I wasn’t sure how robust we wanted our `src/utils/*` to be against misuse.

I wrote a unit test that crashed this, so I made the check verify that it was non-null. All of my `Fatal Internal Error` checks could be removed if you wanted to do so. I found them helpful during the initial implementation phase.

src/utils/get-comment-registry.ts
fbartho: I concur that TypeScript would prevent that. [Responded further here: https://github.com/IanVS/prettier-plugin-sort-imports/pull/84#discussion_r1194435397]

src/utils/get-comment-registry.ts
fbartho: I concur that TypeScript would prevent that. [Responded further here: https://github.com/IanVS/prettier-plugin-sort-imports/pull/84#discussion_r1194435397]

IanVS (approved):

PR #86, opened 2023-05-16
Feature: Add `<BUILTIN_MODULES>` Special Word

fbartho:
- This allows users to specify a gap between built-ins and third-party in a slightly less wonky way, and the plugin's default `importOrder` & related documentation a bit clearer.
- This also allows users to move the built-ins around relative to other imports. (Weakens our previous opinionated decision that built-ins should always be first).

### DIFF
--- a/README.md
+++ b/README.md
@@ -194,8 +194,8 @@ The main way to control the import order and formatting, `importOrder` is a coll
 
 ```js
 [
-    '<BUILTIN_MODULES>', // Built-in modules are always grouped. If you don't configure this special word, it will be injected.
-    '<THIRD_PARTY_MODULES>', // Imports not matched by other special words or regexps -- if you don't configure this special word, it will be injected.
+    '<BUILTIN_MODULES>', // Node.js built-in modules
+    '<THIRD_PARTY_MODULES>', // Imports not matched by other special words or groups.
     '^[.]', // relative imports
 ],
 ```
@@ -247,7 +247,7 @@ import styles from './global.css';
 If you want to group your imports into "chunks" with blank lines between, you can add empty strings like this:
 
 ```json
-"importOrder": ["<BUILT_IN_MODULES>", "", "<THIRD_PARTY_MODULES>", "", "^[.]",]
+"importOrder": ["<BUILT_IN_MODULES>", "", "<THIRD_PARTY_MODULES>", "", "^[.]"]
 ```
 
 e.g.:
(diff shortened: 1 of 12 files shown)


src/utils/get-import-nodes-matched-group.ts
```diff
@@ -5,13 +5,24 @@ import {
...
+const regexCache = new Map<string, RegExp>();
+const cachedRegExp = (regExp: string) => {
+    if (regexCache.has(regExp)) {
+        return regexCache.get(regExp)!;
+    }
+    // Strip <TYPES> when creating regexp
+    const result = new RegExp(regExp.replace(TYPES_SPECIAL_WORD, ''));
+    regexCache.set(regExp, result);
+    return result;
+};
```
IanVS, line 17: Nice little perf improvement here.

src/utils/normalize-import-order-options.ts
```diff
@@ -0,0 +1,33 @@
...
+            importOrder.includes(BUILTIN_MODULES_SPECIAL_WORD) ||
+            importOrder.includes(BUILTIN_MODULES_REGEX_STR)
+        )
+    ) {
+        importOrder = [BUILTIN_MODULES_SPECIAL_WORD, ...importOrder];
+    }
+
+    importOrder = importOrder.map((g) =>
+        g === BUILTIN_MODULES_SPECIAL_WORD ? BUILTIN_MODULES_REGEX_STR : g,
+    );
```
IanVS, line 30: Why do this map here, vs just injecting the regex str on line 25?

IanVS (approved):
Minor suggestions, but nothing blocking.  Nice improvement!

src/utils/normalize-import-order-options.ts
IanVS: Oh right, I was neglecting to consider that we need to support both having the special word in the config and injecting it.  The third-party special word is handled elsewhere.

README.md
```diff
@@ -240,11 +247,9 @@ import styles from './global.css';
 If you want to group your imports into "chunks" with blank lines between, you can add empty strings like this:
 
 ```json
-"importOrder": ["", "<THIRD_PARTY_MODULES>", "", "^[.]",]
+"importOrder": ["<BUILT_IN_MODULES>", "", "<THIRD_PARTY_MODULES>", "", "^[.]",]
```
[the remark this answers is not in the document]
fbartho: I had just noticed these commas when reviewing the readme in night mode last night. Was going to follow with a fix-up PR when I had the chance, but you already did the hard work, haha!

PR #90, opened 2023-05-17
Vue: Sort both script and setup script

IanVS:
Fixes https://github.com/trivago/prettier-plugin-sort-imports/issues/218

### DIFF
--- a/src/preprocessors/vue.ts
+++ b/src/preprocessors/vue.ts
@@ -1,14 +1,31 @@
+import type { parse as Parse } from '@vue/compiler-sfc';
+
 import { PrettierOptions } from '../types';
 import { preprocessor } from './preprocessor';
 
 export function vuePreprocessor(code: string, options: PrettierOptions) {
+    let preprocessedCode = code;
     try {
-        const { parse } = require('@vue/compiler-sfc');
+        const { parse }: { parse: typeof Parse } = require('@vue/compiler-sfc');
         const { descriptor } = parse(code);
-        const content =
-            (descriptor.script ?? descriptor.scriptSetup)?.content ?? code;
 
-        return code.replace(content, `\n${preprocessor(content, options)}\n`);
+        if (descriptor.script) {
+            const { content } = descriptor.script;
+            preprocessedCode = preprocessedCode.replace(
+                content,
+                `\n${preprocessor(content, options)}\n`,
+            );
+        }
+
+        if (descriptor.scriptSetup) {
+            const { content } = descriptor.scriptSetup;
+            preprocessedCode = preprocessedCode.replace(
+                content,
+                `\n${preprocessor(content, options)}\n`,
+            );
+        }
+
+        return preprocessedCode;
     } catch (err) {
         if ((err as NodeJS.ErrnoException).code === 'MODULE_NOT_FOUND') {
             console.warn(
--- a/tests/Vue/__snapshots__/ppsi.spec.ts.snap
+++ b/tests/Vue/__snapshots__/ppsi.spec.ts.snap
@@ -1,5 +1,30 @@
 // Vitest Snapshot v1, https://vitest.dev/guide/snapshot.html
 
+exports[`no-script.vue - vue-verify > no-script.vue 1`] = `
+<template>
+    <router-view />
+</template>
+
+<style lang="less">
+#app {
+height: 100%;
+background-color: inherit;
+}
+</style>
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+<template>
+    <router-view />
+</template>
+
+<style lang="less">
+#app {
+    height: 100%;
+    background-color: inherit;
+}
+</style>
+
+`;
+
 exports[`script-and-setup.vue - vue-verify > script-and-setup.vue 1`] = `
 <script>
 import z from 'z';
--- a/tests/Vue/__snapshots__/ppsi.spec.ts.snap
+++ b/tests/Vue/__snapshots__/ppsi.spec.ts.snap
@@ -48,13 +48,16 @@ function add(a, b) {
 </script>
 
 <script setup>
-import z from "z";
-import threeLevelRelativePath from "../../../threeLevelRelativePath";
-import sameLevelRelativePath from "./sameLevelRelativePath";
 import thirdParty from "third-party";
-import oneLevelRelativePath from "../oneLevelRelativePath";
-import otherthing from "@core/otherthing";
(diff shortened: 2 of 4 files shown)


Tanimodori:
`code.replace` is not safe to use here.

https://github.com/IanVS/prettier-plugin-sort-imports/blob/c2a67662333de61a23af1fead28067e823eef430/src/preprocessors/vue.ts#L85-L88

The `code` here is the whole SFC source code. Despite the fact that up to one `<script>` and one `<script setup>` block can exist in a single SFC, other 3rd blocks may exist in that SFC (like `vue-i18n` adds a new `<i18n>` block). In rare cases `String.prototype.replace` may replace the wrong place producing tricky issues and it is slower than merging by block offsets in [the original fix](https://github.com/trivago/prettier-plugin-sort-imports/pull/219/commits/5ffeef5286a2a10c1db8a74e1ea0052071c6b3c9).

fbartho:
@Tanimodori that sounds like a serious bug! Would you mind filing a PR to at least give us examples, if not fix it?

I’m not a Vue user, so I only sort of understand your situation here. I’m happy to review any PR that fixes this.

Tanimodori:
@fbartho Surely. Give me some time to implement that.

Tanimodori:
@fbartho The fix is blocked by #133. I've opened PR #134 to fix that.

Tanimodori:
Let me introduce some backgrounds and give the examples.

Vue **S**ingle **F**ile **C**omponent, or SFC, is a file that contains blocks which describes various aspects of a reusable component similar to Web Components. Here's a typical structure of SFC:

```vue
<template>
  <div class="my-text">
    Here's the template block that 
    describes how vue render this component
    into HTML DOM
  </div>
</template>

<script setup lang="ts">
  console.log("Here's the script block that");
  console.log("contains the actual code of this component");
</script>

<style lang="less">
// Here is the style block that
// contains the style used in template
.my-text {
  color: red;
}
</style>
```

You can tell that different block use different language and the order of block does not matter. This prettier plugin only cares about `<script>` and `<script setup>` which contains JS/TS code so we use `parse` from `@vue/compiler-sfc` to split the blocks.

Tanimodori:
Here is the problematic example:

```vue
<template>
  <code>
import z from 'z';
import threeLevelRelativePath from "../../../threeLevelRelativePath";
import sameLevelRelativePath from "./sameLevelRelativePath";
import thirdParty from "third-party";
import oneLevelRelativePath from "../oneLevelRelativePath";
import otherthing from "@core/otherthing";
import { defineComponent } from 'vue'
function add(a,b) {
  return a + b;
}

import z from 'z';
import threeLevelRelativePath from "../../../threeLevelRelativePath";
import sameLevelRelativePath from "./sameLevelRelativePath";
import thirdParty from "third-party";
import oneLevelRelativePath from "../oneLevelRelativePath";
import otherthing from "@core/otherthing";
import { defineComponent } from 'vue'
function add(a,b) {
  return a + b;
}
  </code>
</template>

<script>
import z from 'z';
import threeLevelRelativePath from "../../../threeLevelRelativePath";
import sameLevelRelativePath from "./sameLevelRelativePath";
import thirdParty from "third-party";
import oneLevelRelativePath from "../oneLevelRelativePath";
import otherthing from "@core/otherthing";
import { defineComponent } from 'vue'
function add(a,b) {
  return a + b;
}
</script>

<script setup>
import z from 'z';
import threeLevelRelativePath from "../../../threeLevelRelativePath";
import sameLevelRelativePath from "./sameLevelRelativePath";
import thirdParty from "third-party";
import oneLevelRelativePath from "../oneLevelRelativePath";
import otherthing from "@core/otherthing";
import { defineComponent } from 'vue'
function add(a,b) {
  return a + b;
}
</script>
```

The `<template>` block contains the same "code text" in `<script>` and `<script setup>` here! Well the "code text" in `<template>` block only render into HTML DOM which does not work as actual JS/TS code. In some scenario when we want to show the actual code in the component such as demonstrate a specific use case of a component in a component library, this example may be actually exist. Note that 3rd plugins may also add their block into SFC.

So now you can tell the problem. As `String.prototype.replace` only replace the first occurrence and the order of blocks does not matter (some coders prefer writing `<template>` first actually), `code.replace` will replace the context in `<template>` rather than `<script>` and `<script setup>`! Here's the outcome:

```vue
<template>
    <code>
        import thirdParty from "third-party"; import { defineComponent } from
        'vue'; import z from 'z'; import otherthing from "@core/otherthing";
        import threeLevelRelativePath from "../../../threeLevelRelativePath";
        import oneLevelRelativePath from "../oneLevelRelativePath"; import
        sameLevelRelativePath from "./sameLevelRelativePath"; function add(a,b)
        { return a + b; } import thirdParty from "third-party"; import {
        defineComponent } from 'vue'; import z from 'z'; import otherthing from
        "@core/otherthing"; import threeLevelRelativePath from
        "../../../threeLevelRelativePath"; import oneLevelRelativePath from
        "../oneLevelRelativePath"; import sameLevelRelativePath from
        "./sameLevelRelativePath"; function add(a,b) { return a + b; }
    </code>
</template>

<script>
import z from "z";
import threeLevelRelativePath from "../../../threeLevelRelativePath";
import sameLevelRelativePath from "./sameLevelRelativePath";
import thirdParty from "third-party";
import oneLevelRelativePath from "../oneLevelRelativePath";
import otherthing from "@core/otherthing";
import { defineComponent } from "vue";
function add(a, b) {
    return a + b;
}
</script>

<script setup>
import z from "z";
import threeLevelRelativePath from "../../../threeLevelRelativePath";
import sameLevelRelativePath from "./sameLevelRelativePath";
import thirdParty from "third-party";
import oneLevelRelativePath from "../oneLevelRelativePath";
import otherthing from "@core/otherthing";
import { defineComponent } from "vue";
function add(a, b) {
    return a + b;
}
</script>
```

Although the `<code>` is compressed a bit due to the HTML Whitespace Sensitivity settings of `prettier` in this repo's config, you can tell that it has replace the wrong place.

Tanimodori:
As the `parse` tells where a block starts and ends in SFC, the correct and faster way is using that offset to merge blocks like the original fix. The algorithm here is simple here.

1. Find out the block we want, say that we got one `<script>` and one `<script setup>` here. We may have `[0,2]` blocks actually.
2. Sort offsets them by the start one, say that we have `[b1, e1)` and `[b2, e2)`. Blocks may never overlaps.
3. Format the contents of those blocks, say we got string `c1` and `c2`.
4. Merge. We use `[0, b1)` from the original, then `c1`, then `[e1, b2)` from the original, then `c2` and finally `[e2,code.length)`

Tanimodori:
I've submitted PR #135 to fix the problem without touching dependencies and shrimpwarp files anyway. Should be able to merge without conflicts.

Issue #78, opened 2023-05-12
Parser Error with the JSX parser plugin & TypeScript old-style assertions <SomeType>foo


fbartho:
**Your Environment**

-   **Prettier version**: 2.8.8
-   **node version**: v18.11.0
-   **package manager**: yarn@1.22.19
-   **IDE**: VScode up-to-date

**Describe the bug**

I ran 4.0.0-alpha.3 and 4.0.0-alpha.4 on my large monorepo codebase, and found a couple places where the parsing crashed, with a useless error message:

```shell
$ yarn prettier --write […]/devices.service.ts
[error] […]/devices.service.ts: SyntaxError: Unexpected token (48:76)
[error] // and thus follows all the lines of the file.
```

**To Reproduce**

1. Have `jsx` in your `importOrderParserPlugins`
```js
    importOrderParserPlugins : ['typescript', 'jsx', 'decorators-legacy', 'classProperties']
```
2. Have typescript code like so:
```ts
import { b } from "b";
import { a } from "a";

// Minimal replication:
<A>a();

// Less minimal:
// export class SomeClass {
// 	async someMethod(
// 		anArg:string
// 	): Promise<SomeResponseType> {
// 		return {
// 			a: <A>something.a,
// 			b: <B>something.b,
// 		};
// 	}
// }
```

When running inside this repo's test-cases, I get this useful log message:
```
Serialized Error: {
  "clone": "Function<clone>",
  "code": "BABEL_PARSER_SYNTAX_ERROR",
  "codeFrame": "   1 |
   2 | import { b } from "b";
   3 | import { a } from "a";
   4 |
   5 | // Minimal replication:
   6 | <A>a();
   7 |",
  "details": {
    "expected": null,
  },
  "loc": {
    "column": 0,
    "constructor": "Function<Position>",
    "index": 284,
    "line": 19,
  },
  "pos": 284,
  "reasonCode": "UnexpectedToken",
}
```

**Expected behavior**

1. No crash
1. When crashes do occur, I'd like a better error message

**Configuration File (cat .prettierrc, prettier.config.js, .prettier.js)**

```js
module.exports = {
	overrides: [
		{
			// prettier will strip newlines out of package.json files unless you tell it to use the json parser
			files: ["package.json", "**/package.json"],
			options: { parser: "json" },
		},
	],
	printWidth: 100,
	semi: true,
	singleQuote: false, // Single quotes are common in text-strings
[...]

PR #91, opened 2023-05-17
Use correct parser plugins for TS and Vue

IanVS:
This fixes a few issues with parser plugins.

It fixes https://github.com/IanVS/prettier-plugin-sort-imports/issues/78 by removing the `jsx` parser plugin when the file is named `.ts` (typescript requires the use of `.tsx` extension for jsx).

It also applies similar logic for .vue files, ensuring that the correct parser plugins are used, with similar logic to the vue compiler itself.

### DIFF
--- a/src/preprocessors/preprocessor.ts
+++ b/src/preprocessors/preprocessor.ts
@@ -10,7 +10,7 @@ import { getExperimentalParserPlugins } from '../utils/get-experimental-parser-p
 import { getSortedNodes } from '../utils/get-sorted-nodes';
 
 export function preprocessor(code: string, options: PrettierOptions): string {
-    const { importOrderParserPlugins, importOrder } = options;
+    const { importOrderParserPlugins, importOrder, filepath } = options;
     let { importOrderTypeScriptVersion } = options;
     const isTSSemverValid = semver.valid(importOrderTypeScriptVersion);
 
@@ -29,10 +29,15 @@ export function preprocessor(code: string, options: PrettierOptions): string {
         : true;
 
     const allOriginalImportNodes: ImportDeclaration[] = [];
+    let plugins = getExperimentalParserPlugins(importOrderParserPlugins);
+    // Do not inject jsx plugin for non-jsx ts files
+    if (filepath.endsWith('.ts')) {
+        plugins = plugins.filter((p) => p !== 'jsx');
+    }
     const parserOptions: ParserOptions = {
         sourceType: 'module',
         attachComment: true,
-        plugins: getExperimentalParserPlugins(importOrderParserPlugins),
+        plugins,
     };
 
     // Disable importOrderCombineTypeAndValueImports if typescript is not set to a version that supports it
--- a/src/preprocessors/vue.ts
+++ b/src/preprocessors/vue.ts
@@ -75,13 +75,6 @@ function sortScript(
         if (!hasPlugin(newPlugins, 'typescript')) {
             newPlugins.push('typescript');
         }
-
-        if (
-            !hasPlugin(newPlugins, 'decorators') &&
-            !hasPlugin(newPlugins, 'decorators-legacy')
-        ) {
-            newPlugins.push('decorators-legacy');
-        }
     }
 
     const adjustedOptions = {
--- a/src/preprocessors/vue.ts
+++ b/src/preprocessors/vue.ts
@@ -1,6 +1,8 @@
 import type { parse as Parse } from '@vue/compiler-sfc';
 
+import { ImportOrderParserPlugin } from '../../types';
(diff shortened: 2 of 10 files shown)


src/preprocessors/preprocessor.ts
```diff
@@ -29,10 +29,15 @@ export function preprocessor(code: string, options: PrettierOptions): string {
         : true;
 
     const allOriginalImportNodes: ImportDeclaration[] = [];
+    let plugins = getExperimentalParserPlugins(importOrderParserPlugins);
+    // Do not inject jsx plugin for non-jsx ts files
+    if (filepath.endsWith('.ts')) {
+        plugins = plugins.filter((p) => p !== 'jsx');
```
fbartho, line 35: Shouldn’t this use a fancier logic in case people pass options to the jsx plugin? (Or does it have no options?)

src/utils/get-experimental-parser-plugins.ts
```diff
@@ -32,3 +32,37 @@ export const getExperimentalParserPlugins = (
...
+    importOrderParserPlugins: string[],
+    pluginName: string,
+): boolean => {
+    for (const pluginNameOrJson of importOrderParserPlugins) {
+        const isParserPluginWithOptions = pluginNameOrJson.startsWith('[');
+        let plugin;
+
+        if (isParserPluginWithOptions) {
+            try {
+                plugin = JSON.parse(
```
fbartho, line 53: I think you forgot to select the first node in the Tuple `[“pluginName”, …]` needs a `[0]` somewhere near here right?

fbartho (changes requested):
Commented on the two things I had concerns around!

src/preprocessors/preprocessor.ts
IanVS: jsx has no options.  

```ts
// babel-parser.d.ts

type ParserPluginWithOptions =
  | ["decorators", DecoratorsPluginOptions]
  | ["estree", { classFeatures?: boolean }]
  // @deprecated
  | ["moduleAttributes", { version: "may-2020" }]
  | ["pipelineOperator", PipelineOperatorPluginOptions]
  | ["recordAndTuple", RecordAndTuplePluginOptions]
  | ["flow", FlowPluginOptions]
  | ["typescript", TypeScriptPluginOptions];
```

src/utils/get-experimental-parser-plugins.ts
IanVS: ah, yeah good call

src/preprocessors/vue.ts
```diff
@@ -35,3 +37,60 @@ export function vuePreprocessor(code: string, options: PrettierOptions) {
...
+            newPlugins.push('typescript');
+        }
+
+        if (
+            !hasPlugin(newPlugins, 'decorators') &&
+            !hasPlugin(newPlugins, 'decorators-legacy')
+        ) {
+            newPlugins.push('decorators-legacy');
+        }
+    }
```
fbartho, line 85: I’m uncomfortable with injecting `decorators`, because I’ve seen people fork every plugin. I worry that we’ll always be injecting decorators-legacy.

- https://babeljs.io/docs/babel-plugin-proposal-decorators

I’d be less uncomfortable if you were to shift this logic into the conditional-injection of the `‘typescript’` plugin. (Move the latter if-block to line 77).

I’d be least uncomfortable if this logic only happened if users omitted `importOrderParserPlugins` entirely. Then we definitely have permission to do automatic behavior. — If users are passing us plugins, I worry that manipulating it could only lead to more, painful errors.

fbartho (commented):
If this logic *only* kicked in if users omit the `importOrderParserPlugins` options, or if they passed `“auto-detect”` then I’d be a total thumbs up!

src/preprocessors/vue.ts
IanVS: One problem we have is that we always inject `typescript` and `jsx`, and there's no way to know whether it was us or the user who did it.  I'd like to get to the point where we can remove those defaults and be smart about when we default to various plugins, but I'm not all the way there yet.  I'll see what it would take to do that as part of 4.0.

As for the decorators specifically, that was something I copied straight from vue (see https://github.com/vuejs/vue/commit/326d24a4e42b4f4fd243b36037e88b71963d963d).  I'm not a vue user, so I'm not sure why it's important to include a decorator plugin, but I think it might be related to the older class-based component definition in Vue 2: https://v2.vuejs.org/v2/guide/typescript.html#Class-Style-Vue-Components.

So, I'm happy to remove the injection of the decorators plugin.  Are you okay with the rest of it staying as-is?

fbartho (approved):

PR #92, opened 2023-05-17
Feature: Allow Explicit Separator after Top-of-file-comments

fbartho:
Primarily, this makes it possible for people to declare that there should be a gap after top-of-file-comments (before other imports).

Additionally:
- Centralizes plugin options parsing so it only happens once
- I think it fixes a regression that was introduced in `4.0.0-alpha.?` around `tests/ImportsSeparatedIntoSideEffectGroups`
- Cleans up some more trailing commas in the readme-json-options


src/utils/get-sorted-nodes-by-import-order.ts
```diff
@@ -14,12 +19,19 @@ import {
...
-    const importOrder = normalizeImportOrderOption(options.importOrder);
+    if (
+        process.env.NODE_ENV === 'test' &&
+        JSON.stringify(importOrder) !==
+            JSON.stringify(testingOnly.normalizeImportOrderOption(importOrder))
+    ) {
+        throw new Error(
+            'API Misuse: getSortedNodesByImportOrder::importOrder option already should be normalized.',
+        );
+    }
```
IanVS, line 34: If we think this could be a point of confusion, maybe it's worth renaming the option after we process it?  Then we could rely on TypeScript vs runtime (testtime) checks.

README.md
```diff
@@ -299,6 +300,31 @@ import icon from '@assets/icon';
...
 
+##### 6. Enforce a blank line after top of file comments
+
+If you have pragma-comments at the top of file, or you have boilerplate copyright announcements, you may be interested in separating that content from your code imports. Explicitly providing `"<BUILTIN_MODULES>", "<THIRD_PARTY_MODULES>"` in your list will allow you to add a separator before them!
+
+```json
+"importOrder": [
+    "", // Include the separator here! And also provide "<BUILTIN_MODULES>", "<THIRD_PARTY_MODULES>" later
+    "<BUILTIN_MODULES>",     // This must be present in the list somewhere
+    "<THIRD_PARTY_MODULES>", // This must be present in the list somewhere
```
IanVS, line 311: What's the reason for the limitation that both builtins and third party need to be in the list for this to work?  Because if they're not there it's ambiguous what we should do, whether we should inject those magic strings before or after the newline?

tests/ImportsSeparatedIntoSideEffectGroups/__snapshots__/ppsi.spec.ts.snap
```diff
@@ -74,6 +76,7 @@ import anotherSameLevelRelativePath1 from "./anotherSameLevelRelativePath1";
 import sameLevelRelativePath1 from "./sameLevelRelativePath1";
 
 import "side-effect-x";
+
```
IanVS, line 79: Hmmm, why are these newlines being added?  I tried to get rid of it, and the only way I could was to remove _all_ separators from importOrder.  Is that intentional?  I thought we were only adding blank lines above the side-effect imports when we found a custom separator.

README.md
```diff
@@ -299,6 +300,31 @@ import icon from '@assets/icon';
...
+```
+
+e.g.:
+
+```ts
+/**
+ * @prettier
+ */
+
+import { debounce, reduce } from 'lodash';
```
fbartho, line 322:
```suggestion
import { promises } from 'fs';
```

src/utils/normalize-plugin-options.ts
```diff
@@ -0,0 +1,103 @@
...
+        importOrderCombineTypeAndValueImports = false;
+    }
+    return {
+        importOrder,
+        importOrderCombineTypeAndValueImports,
+        hasAnyCustomGroupSeparatorsInImportOrder: importOrder.some(
+            isCustomGroupSeparator,
+        ),
+        // Now that the regex for <BUILTIN_MODULES> is present, we can check if the user
+        // configured a separator before z<BUILTIN_MODULES>
```
fbartho, line 96:
```suggestion
        // configured a separator before <BUILTIN_MODULES>
```

IanVS (commented):
Still going through it, but wanted to leave a few comments as I go.

README.md
fbartho: It’s a consequence of how we normalize the importOrder:
1. Check if `<THIRD_PARTY_MODULES>` is present, if not: inject it at front
2. Check if `<BUILTIN_MODULES>` is present, if not: inject at front
3. Check if first element is empty separator: if yes, set the flag `providesGapAfterTopOfFileComments`

Since we only know if you meant “after file-comments” if the first element is a separator, we need you to insert the special words so that you can put a separator before them.

I’m not sure how we could change normalization without having worse consequences?

src/types.ts
```diff
@@ -46,13 +47,28 @@ export type SomeSpecifier =
     | ImportNamespaceSpecifier;
 export type ImportRelated = ImportOrLine | SomeSpecifier;
 
+export interface InspectedAndNormalizedOptions {
```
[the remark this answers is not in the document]
fbartho: I’m fine with either name! Want me to rename it to `ExtendedOptions`?

tests/ImportsSeparatedIntoSideEffectGroups/__snapshots__/ppsi.spec.ts.snap
fbartho: These new lines were supposed to be there, according to the comment in tests/ImportsSeparatedIntoSideEffects/ppsi.spec.ts there was a comment:
```
run_spec(__dirname, ['typescript'], {
    importOrder: [
        '^@core/(.*)$',
        '^@server/(.*)',
        '^@ui/(.*)$',
        '^[./]',
        '', // This adds a newline before side effect groups
    ],
});
```
I’m pretty sure this change is just me restoring that expectation!

src/utils/__tests__/remove-nodes-from-original-code.spec.ts
```diff
@@ -23,7 +24,7 @@ test('it should remove nodes from the original code', () => {
     const ast = babelParser(code, { sourceType: 'module' });
     const importNodes = getImportNodes(code);
     const sortedNodes = getSortedNodes(importNodes, {
-        importOrder: [],
+        importOrder: testingOnly.normalizeImportOrderOption([]),
```
IanVS, line 27: Instead of exporting this for testing, we could consider "pre-normalizing" all of the arrays used here in these tests (i.e. making sure they all specify builtins and third party).  Don't feel strongly about it, only that exporting code only to use in testing feels like a smell.

src/utils/get-sorted-nodes-by-import-order.ts
fbartho: This is only matters when people are directly calling this utility function, eg. From our spec files?

I threw this error here and then was able to track down all the test-files that needed corrections for moving the importOrder normalization to somewhere else.

I’ve tagged this if-block so it only executes in `test` but if you prefer, we could delete it.

src/utils/__tests__/get-sorted-nodes-by-import-order.spec.ts
```diff
@@ -179,8 +182,13 @@ test('it returns all sorted nodes with builtin specifiers at the top', () => {
 test('it returns all sorted nodes with custom third party modules and builtins at top', () => {
     const result = getImportNodes(code);
     const sorted = getSortedNodesByImportOrder(result, {
-        importOrder: ['^a$', '<THIRD_PARTY_MODULES>', '^t$', '^k$', '^[./]'],
-        importOrderCombineTypeAndValueImports: true,
```
IanVS, line 183: I think this is lost if we don't specify a `importOrderTypeScriptVersion`, right?  I don't think you meant to change that behavior in this test.

src/utils/normalize-plugin-options.ts
```diff
@@ -31,3 +35,64 @@ export function normalizeImportOrderOption(
...
+
+/**
+ * isCustomGroupSeparator checks if the provided pattern is intended to be used
+ * as an import separator, rather than an actual group of imports.
+ */
+export function isCustomGroupSeparator(pattern?: string) {
+    return pattern?.trim() === '';
+}
+
+export function examineAndNormalizePluginOptions(
```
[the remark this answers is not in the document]
fbartho:
```suggestion
/** 
 * Verifies that our special words that must always be there are present on importOrder
 * Verifies that parser plugins are inferred correctly for certain file extensions.
 *
 * Configures certain behavior flags such as
 *  - when to use certain typescript syntax
 *  - when to inject blank lines after top-of-file comments
 *  - when to inject blank lines around groups / side-effect nodes.
 */
export function examineAndNormalizePluginOptions(
```

README.md
IanVS: My only idea would be to check if the first element is a separator, and if so, inject third party and builtins under it.  But like I said, that might not be what the user wanted, because maybe they wanted a gap between those and their own code.  But, if they do, then they could add those special words explicitly, and honestly if that's their intent then specifying them explicitly is preferred for understandability.  

So, I have a minor preference to adjusting as I suggested, and making a blank line at the top always mean "space after top-of-file comments".  That just feels a bit easier to explain/understand.  What do you think?

tests/ImportsSeparatedIntoSideEffectGroups/__snapshots__/ppsi.spec.ts.snap
IanVS: But this is a newline _after_ the side-effect node, not between side-effect groups, right?  That doesn't seem to be what that comment was about.

src/types.ts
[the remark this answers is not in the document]
IanVS: Sorry to bikeshed, but yeah if you don't mind I'd prefer that.

src/utils/get-comment-registry.ts
```diff
@@ -422,9 +428,14 @@ export function attachCommentsToOutputNodes(
                     line: firstImport.loc?.end.line + commentHeight,
                 },
             };
+
             const moveDist =
                 originalLoc.start.line - newFirstImport.loc.start.line;
 
+            if (provideGapAfterTopOfFileComments) {
```
IanVS, line 435: I guess we also need to make sure that the top-of-file comment isn't a `prettier-ignore` comment.

IanVS (commented):
OK I think I'm on board with this, and don't have further comments.  Let me know when you'd like me to take another look.

src/utils/__tests__/remove-nodes-from-original-code.spec.ts
fbartho: - pre-normalizing might let subsequent developers think those fields are optional.
- I guess I wanted to preserve the developer intent: "I intend to provide an empty array to the plugin" -- and the normalization is an implementation detail of that?

tests/ImportsSeparatedIntoSideEffectGroups/__snapshots__/ppsi.spec.ts.snap
fbartho: I guess I'm confused now. I'm seeing it as a newline before *any* chunk, in this case the following chunk.

I think it was weird to have a special meaning to the last slot, because side-effect imports aren't "last". We just create a chunk before them even if there are no imports in that chunk, but I guess some people would consider the side-effect chunks the last entry in a chunk?

idk. I'm confused on this now! The sample looks reasonable to me, but I'm not sure how I would configure things around side-effects if I didn't know the implementation.

IanVS: OK, I was thinking about side-effect nodes _starting_ a chunk, but in reality they _become_ a chunk.  Which I guess makes sense.  So, side-effect imports will be broken out on their own as long as there's a custom group separator somewhere in the `importOrder`.  That keeps them from looking like they're part of either the group before or the group after, each of which is sorted according to the `importOrder`.  So, I think the change here is correct after all.  Thanks.

docs/MIGRATION.md
```diff
@@ -22,7 +22,7 @@ For example:
 
 ```js
     "importOrder": [
-        "", // This emptry group at the start will add separators for side-effect imports and node.js built-in modules
+        "", // If you want a gap at the top after top-of-file-comments, put a separator here!
```
IanVS, line 25: We might want to include `"<BUILTIN_MODULES>", ""` here too, to really replicate `importOrderSeparation`

IanVS, line 25:
```suggestion
        "", // If you want a gap at the top after top-of-file-comments, put a separator here!
        "<BUILTIN_MODULES>",
        "",
```

IanVS (approved):

README.md
fbartho: (Implemented, resolving discussion)

docs/MIGRATION.md
fbartho: Done!

fbartho: Done!

PR #95, opened 2023-05-19
Inject required groups after leading separator if present

fbartho:
Created as a separate PR to review if I'm missing anything


tests/ImportOrderBuiltinModulesToTop/__snapshots__/ppsi.spec.ts.snap
```diff
@@ -6,7 +6,6 @@ import thirdParty from 'third-party';
 import fs from "fs"
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 import fs from "fs";
-
```
IanVS, line 9: Note, to keep this newline, we'd need to specify `<BUILTIN_MODULES>` in the `importOrder`, but the point of this test is to show that we do that automatically, so we'll just remove the newline and it's fine.

IanVS (approved):

Issue #96, opened 2023-05-21
Does the plugin have Astro compatibility?


RodrigoTomeES:
Hi,

Does the plugin have Astro compatibility?

Thanks!

IanVS:
Yes, I use it in my own Astro project.

RodrigoTomeES:
@IanVS I mentioned it because I recently returned to a project using the latest version of Astro, and it stopped working. The imports are no longer being sorted, and since it's not listed in the compatibility table, I thought that might be the issue.

IanVS:
@RodrigoTomeES can you share your `.prettierrc` or even better a reproduction?

RodrigoTomeES:
@IanVS sure, here you have. thanks!

```json
{
  "semi": true,
  "singleQuote": true,
  "endOfLine": "lf",
  "tabWidth": 2,
  "useTabs": false,
  "printWidth": 80,
  "overrides": [
    {
      "files": ["**/*.css"],
      "options": {
        "singleQuote": false
      }
    }
  ],
  "plugins": [
    "prettier-stylelint",
    "prettier-plugin-astro",
    "@ianvs/prettier-plugin-sort-imports",
    "prettier-plugin-tailwindcss"
  ],
  "importOrder": [
    "<THIRD_PARTY_MODULES>",
    "",
    "^@/atoms/(.*)$",
    "^@/molecules/(.*)$",
    "^@/organisms/(.*)$",
    "^@/templates/(.*)$",
    "^@/pages/(.*)$",
    "^@/layouts/(.*)$",
    "",
    "^@/(.*)$",
    "^[./]",
    "",
    "<TYPES>",
    "<TYPES>^[./]"
  ],
  "importOrderMergeDuplicateImports": true,
  "pluginSearchDirs": false
}
```

IanVS:
It could be because you're using `prettier-plugin-tailwindcss`.  Recent versions do include this plugin in its compatibility list, so make sure you're on the latest there.

I upgraded my project to Astro 2.5.1, and was still able to sort correctly.  I can look into it a bit more if you're willing to put together a small reproduction and share it.

RodrigoTomeES:
@IanVS mmm I have all my dependencies updated, look:

```bash
$ yarn outdated
➤ YN0000: ┌ Checking for outdated dependencies
➤ YN0000: └ Completed in 2s 917ms

➤ YN0000: ✨ All your dependencies are up to date!
[...]

PR #97, opened 2023-05-22
Support babel-ts parser

IanVS:
Restores compatibility with the latest version of `prettier-plugin-astro`.

Fixes https://github.com/IanVS/prettier-plugin-sort-imports/issues/96.

To confirm, I made this change in the `node_modules` of the project in the above issue, and confirmed that sorting began working correctly again.


fbartho (commented):
👍

PR #99, opened 2023-05-28
Fix cli in Install section

torn4dom4n:
I updated the command script in the Install section.


fbartho (approved):

IanVS:
Thanks so much!  Looks like I deleted a bit too much when I removed the `@next` recently, oops.

PR #101, opened 2023-05-30
Fix dollar sign ($) issue in Vue SFC

### LINKED ISSUE #100: [V4] Dollar sign ($) in Vue SFC script setup throws `SyntaxError`
**Your Environment**

-   **Prettier version**: 2.8.7
-   **node version**: v18.16.0
-   **package manager**: npm@9.5.1

**Describe the bug**

Using a dollar sign `$` inside the vue SFC script setup throws an exception and doesn't format the file.

**To Reproduce**

Try to format a Vue SFC file with a single quoted `$` inside `<script lang="ts" setup>`

**ExampleOne.vue**
```vue
<script lang="ts" setup>
const dollarSign = '$';
</script>

<template>
  <div>
    {{ dollarSign }}
  </div>
</template>

```

```js
module.exports = {
    singleQuote: true,
    pluginSearchDirs: false,
    plugins: ["@ianvs/prettier-plugin-sort-imports"],
};
```

**Error log**
```
❯ npx prettier --write *.vue
ExampleOne.vue
[error] ExampleOne.vue: SyntaxError: Unexpected closing tag "script". It may happen when the tag has already been closed by another tag. For more info see https://www.w3.org/TR/html5/syntax.html#closing-elements-that-have-implied-end-tags (12:1)
[error]   10 | ;
[error]   11 |
[error] > 12 | </script>
[error]      | ^^^^^^^^^
[error]   13 |
[error]   14 | <template>
[error]   15 |   <div>
```

istiak-tridip:
Fixes #100

When replacing a string if the provided replacement is a string it will replace the original string using pattern matching. In the vue preprocessor, the `$` in the code is seen as a regex pattern. Passing a function as a replacement escapes that issue.

### DIFF
--- a/src/preprocessors/vue.ts
+++ b/src/preprocessors/vue.ts
@@ -84,6 +84,6 @@ function sortScript(
 
     return code.replace(
         content,
-        `\n${preprocessor(content, adjustedOptions)}\n`,
+        () => `\n${preprocessor(content, adjustedOptions)}\n`,
     );
 }
--- a/tests/Vue/__snapshots__/ppsi.spec.ts.snap
+++ b/tests/Vue/__snapshots__/ppsi.spec.ts.snap
@@ -149,6 +149,10 @@ import { defineComponent } from 'vue'
 function add(a,b) {
   return a + b;
 }
+
+const number = 500;
+const dollarSign = '$';
+const dollarPrefixed =\`\\$\${number}\`;
 </script>
 
 <template>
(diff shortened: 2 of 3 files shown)


IanVS (approved):

IanVS:
Oh beautiful, thanks for digging in and submitting a fix + snapshot test.

fbartho (commented):
I admit I don’t understand why this works?

IanVS:
@fbartho it sounds like we were hitting this: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replace#specifying_a_string_as_the_replacement

fbartho:
Ah! I didn't know there was extra special escapes in that case. Thanks for replying!

PR #105, opened 2023-06-01
Bugfix: handle case where `filepath` isn't available due to programatic access of prettier

sitch:
add option chain


IanVS:
Thanks for the PR, but this change was already made in https://github.com/IanVS/prettier-plugin-sort-imports/pull/104.

fbartho:
Thanks for contributing @sitch -- we actually had a preceding PR #104 with almost the same fix as you contributed (Some additional typescript shenanigans).

@IanVS is will soon release a patch release with the fix in it.

Thanks again!

PR #123, opened 2023-09-18
Update semver

useafterfree:
https://github.com/advisories/GHSA-c2qf-rxjj-qqgw

### DIFF
--- a/package.json
+++ b/package.json
@@ -58,7 +58,7 @@
         "@babel/parser": "^7.21.8",
         "@babel/traverse": "^7.21.5",
         "@babel/types": "^7.21.5",
-        "semver": "^7.5.0"
+        "semver": "^7.5.2"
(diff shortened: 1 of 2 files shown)


IanVS:
Thanks, please run `yarn install` and check in the new updated `yarn.lock`.

Also, note that this issue doesn't really impact us since we don't send untrusted user input through the semver package, we only examine your config file with it.

useafterfree:
> Thanks, please run `yarn install` and check in the new updated `yarn.lock`?
> 
> Also, note that this issue doesn't really impact us since we don't send untrusted user input through the semver package, we only examine your config file with it.

Done!

We get flagged in container scans for this, despite it not being in the codepath

IanVS:
Understood.  Thanks for the PR!

PR #125, opened 2023-09-25
Fix #124 - type import specifier dupe

### LINKED ISSUE #124: Multiple inline-style type imports from the same file adds a side-effect import for the same file
**Your Environment**

-   **Prettier version**: 3.03
-   **node version**: 20.2.0
-   **package manager**: npm@9.6.6
-   **IDE**: CLI

**Describe the bug**

When running prettier on a file that has multiple inline type imports from the same file:

```typescript
import { type x, type y } from "./myFile";
```

the plugin will add another side-effect import for the same file:

```typescript
import "myFile";
import type { x, y } from "./myFile";
```

**To Reproduce**

```typescript
// typeImportTest.types.ts
export type TFirstTest = "first";
export type TSecondTest = "second";
export type TThirdTest = "third";
```

```typescript
// typeImportTest.ts
import { type TFirstTest, type TSecondTest } from "./typeImportTest.types";

const first: TFirstTest = "first";
const second: TSecondTest = "second";
```

Run `prettier -w typeImportTest.ts`

**Expected behaviour**

File should remain unchanged.

**Actual behaviour**

```typescript
import "./typeImportTest.types";
import type { TFirstTest, TSecondTest } from "./typeImportTest.types";

const first: TFirstTest = "first";
const second: TSecondTest = "second";
```

**Configuration File (.prettierrc.js)**
```javascript
module.exports = {
    singleAttributePerLine: true,
    tabWidth: 4,
    plugins: ["@ianvs/prettier-plugin-sort-imports"],
    importOrderTypeScriptVersion: "^4.9.3",
    importOrder: [
        "<TYPES>",
    ],
};
```

**Additional thoughts**

It looks like the plugin is attempting to find non-type imports in the destructuring-style import and break them out into a separate line, since if you import a variable from the file as well:
```typescript
import { type TFirstTest, type TSecondTest, variable } from "./typeImportTest.types";
```
then the resulting output is valid:
```typescript
import { variable } from "./typeImportTest.types";
import type { TFirstTest, TSecondTest } from "./typeImportTest.types";
```

So it looks like the fix would be to check whether there are any non-type imports extracted from the import group before appending another import.

AlexJDG:
Proposed fix for the bug ticket I opened (Fixes https://github.com/IanVS/prettier-plugin-sort-imports/issues/124)

- Prevented type import exploding when no value imports are present
- Added test coverage to cover this edge-case

### DIFF
--- a/src/utils/__tests__/explode-type-and-value-specifiers.spec.ts
+++ b/src/utils/__tests__/explode-type-and-value-specifiers.spec.ts
@@ -82,6 +82,20 @@ test('it should return named type imports unchanged', () => {
     );
 });
 
+test('it should return inline named type imports unchanged', () => {
+    const code = `import { type NamedType1, type NamedType2 } from './source';`;
+    const importNodes = getImportNodes(code, { plugins: ['typescript'] });
+    const explodedNodes = explodeTypeAndValueSpecifiers(importNodes);
+    const formatted = getCodeFromAst({
+        nodesToOutput: explodedNodes,
+        originalCode: code,
+        directives: [],
+    });
+    expect(formatted).toEqual(
+        `import { type NamedType1, type NamedType2 } from './source';`,
+    );
+});
+
 test('it should separate named type and value imports', () => {
     const code = `import { named, type NamedType } from './source';`;
     const importNodes = getImportNodes(code, { plugins: ['typescript'] });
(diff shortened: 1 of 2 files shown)


IanVS (approved):
Nice catch, and thanks for the fix + test!

PR #135, opened 2023-10-12
fix: vue preprocessor may replace wrong code

Tanimodori:
Fix problem mentioned in #90 without changing dependency and shrimpwarp files.

See https://github.com/IanVS/prettier-plugin-sort-imports/pull/90#issuecomment-1758812882 for more details and explanation.


IanVS (approved):
Great, thanks so much for the PR, it looks good to me.  And thanks for the test case and the thorough explanation in your comments.

PR #137, opened 2023-10-23
chore(deps): bump @babel/traverse to avoid audit warning

pauliesnug:
more info: https://github.com/advisories/GHSA-67hx-6x53-jw92

### DIFF
--- a/package.json
+++ b/package.json
@@ -56,7 +56,7 @@
         "@babel/core": "^7.21.8",
         "@babel/generator": "^7.21.5",
         "@babel/parser": "^7.21.8",
-        "@babel/traverse": "^7.21.5",
+        "@babel/traverse": "^7.23.2",
         "@babel/types": "^7.21.5",
         "semver": "^7.5.2"
     },


IanVS:
Thanks for the PR

IanVS:
Note that if you are getting this warning, you can uninstall and reinstall this plugin, or clear and recreate your lock file. Since the version range we depend on includes the good version

echeequorum:
does this change need to be published in a new version of the package? uninstalling and reinstalling still appears to pull back a package.json with the previous version identified.

IanVS:
>uninstalling and reinstalling still appears to pull back a package.json with the previous version identified.

Can you explain what you mean?  You should do something like `npm list @babel/traverse` to see what versions are installed in your project, and where they are coming from.

echeequorum:
thanks for a speedy reply, so installing this as a fresh dependency, so uninstalling it, having it removed from the lockfile then reinstalling it still generates this in the lockfile:

```
    "node_modules/@ianvs/prettier-plugin-sort-imports": {
      "version": "4.1.1",
      "resolved": "https://registry.npmjs.org/@ianvs/prettier-plugin-sort-imports/-/prettier-plugin-sort-imports-4.1.1.tgz",
      "integrity": "sha512-kJhXq63ngpTQ2dxgf5GasbPJWsJA3LgoOdd7WGhpUSzLgLgI4IsIzYkbJf9kmpOHe7Vdm/o3PcRA3jmizXUuAQ==",
      "dev": true,
      "dependencies": {
        "@babel/core": "^7.21.8",
        "@babel/generator": "^7.21.5",
        "@babel/parser": "^7.21.8",
        "@babel/traverse": "^7.21.5",
        "@babel/types": "^7.21.5",
        "semver": "^7.5.2"
      },
      "peerDependencies": {
        "@vue/compiler-sfc": ">=3.0.0",
        "prettier": "2 || 3"
      },
      "peerDependenciesMeta": {
        "@vue/compiler-sfc": {
          "optional": true
        }
      }
    },
```

and I was wondering if that is the case because even though you have updated the package.json file in the repo here, it's not been published in a 4.1.2 version as a patch.

IanVS:
>@babel/traverse is still "^7.21.5"

This is a semver range, which means that the highest version of 7.x.x available will be installed, which is why I said you should check what's actually being installed.  

I plan on publishing a new version of this package soon, but in the meantime, you should not be getting a problematic version of the dependency.  Are you still getting audit warnings?

Issue #142, opened 2024-01-30
`node:test` is not detected as `<BUILTIN_MODULES>`


IsaiahByDayah:
**Your Environment**

-   **Prettier version**: 3.0.3
-   **node version**: 20.11.0
-   **package manager**: pnpm@8
-   **IDE**: VScode

**Describe the bug**

Imports from `node:test` are not properly grouped with other builtin modules

**To Reproduce**

Try to import from `node:test` and notice that it is not included with other `node:*` imports

**Expected behavior**

imports from `node:test` would be grouped with the other builtin modules

**Screenshots, code sample, etc**

https://github.com/IanVS/prettier-plugin-sort-imports/assets/6565265/a7dee908-d49a-4e73-b461-37457c20b38c

**Configuration File (cat .prettierrc, prettier.config.js, .prettier.js)**

```
importOrder: [
    "",
    "<BUILTIN_MODULES>", // Node.js built-in modules
    "",
    "<THIRD_PARTY_MODULES>", // Imports not matched by other special words or groups.
    "",
    "^[.]", // relative imports
  ],
```

**Error log**

N/A

**Contribute to @ianvs/prettier-plugin-sort-imports**

-   [x] I'm willing to fix this bug 🥇

IsaiahByDayah:
Digging into how `<BUILTIN_MODULES>` determines what the builtin module are, it looks to use the exported `builtinModules` array from `import { builtinModules } from "module"` in [`constants.ts`](https://github.com/IanVS/prettier-plugin-sort-imports/blob/main/src/constants.ts#L1).

Following [this open node issue](https://github.com/nodejs/node/issues/42785), it seems like the node maintainers can not (or will not?) add `"test"` to the exported `builtinModules` array (the issue is still open so maybe it will be added one day?). Regardless, because [the regex for determining builtin modules](https://github.com/IanVS/prettier-plugin-sort-imports/blob/main/src/constants.ts#L26) assumes the `builtinModules` array contains _all_ the actual builtin modules, `"node:test"` is not captured in the regex and gets split from the other modules

**Suggestion**

My suggestion to fix would be to explicitly add a check for `"node:test"` to `BUILTIN_MODULES_REGEX_STR`, or at least allow for a way to capture modules that aren't a part of the exported `builtinModules` array (as it seems that can be out of date / not all encompassing)

IanVS:
[...]

PR #143, opened 2024-01-31
Fix: `node:test` not grouped with `<BUILTIN_MODULES>`

IsaiahByDayah:
Following discussion in #142 , this adds an explicit check for `"node:test"` to `BUILTIN_MODULES_REGEX_STR` in order to properly group it with other builtin modules.

Test covers:
- `"node:test"` correctly gets grouped in with other builtin node modules
- `"node:test"` is correctly sorted among other builtin node modules
- `"*node:test"` is NOT grouped in with builtin modules
- `"node:test*"` is NOT grouped in with builtin modules
- `"node"` is NOT grouped in with builtin modules
- `"test"` is NOT grouped in with builtin modules

Fixes https://github.com/IanVS/prettier-plugin-sort-imports/issues/142

### DIFF
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -25,7 +25,7 @@ export const mergeableImportFlavors = [
 
 export const BUILTIN_MODULES_REGEX_STR = `(^(?:node:)?(?:${builtinModules.join(
     '|',
-)})$)|(^(?:node:test)$)`; // NOTE: 'test' is not included in `builtinModules` so explicitly check for "node:test"
+)})$)|(^node:test$)`; // NOTE: 'test' is not included in `builtinModules` so explicitly check for "node:test"
 
 export const BUILTIN_MODULES_SPECIAL_WORD = '<BUILTIN_MODULES>';
 /**
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -23,9 +23,9 @@ export const mergeableImportFlavors = [
     importFlavorType,
 ] as const;
 
-export const BUILTIN_MODULES_REGEX_STR = `^(?:node:)?(?:${builtinModules.join(
+export const BUILTIN_MODULES_REGEX_STR = `(^(?:node:)?(?:${builtinModules.join(
     '|',
-)})$`;
+)})$)|(^(?:node:test)$)`; // NOTE: 'test' is not included in `builtinModules` so explicitly check for "node:test"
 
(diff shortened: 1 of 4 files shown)


fbartho (approved):
Looks good! — Thanks for the clean comment.

src/constants.ts
```diff
@@ -23,9 +23,9 @@ export const mergeableImportFlavors = [
     importFlavorType,
 ] as const;
 
-export const BUILTIN_MODULES_REGEX_STR = `^(?:node:)?(?:${builtinModules.join(
+export const BUILTIN_MODULES_REGEX_STR = `(^(?:node:)?(?:${builtinModules.join(
     '|',
-)})$`;
+)})$)|(^(?:node:test)$)`; // NOTE: 'test' is not included in `builtinModules` so explicitly check for "node:test"
```
fbartho, line 28:
```suggestion
)})$)|(^node:test$)`; // NOTE: 'test' is not included in `builtinModules` so explicitly check for "node:test"
```

I think this might be a simpler regexp?

fbartho (changes requested):
Simplification suggestion, untested

src/constants.ts
IsaiahByDayah: Ah, yup, that seems to work. Updating! (RegEx is one of my gaps I still need to work on 😅)

fbartho (approved):

src/constants.ts
```diff
@@ -23,9 +23,9 @@ export const mergeableImportFlavors = [
     importFlavorType,
 ] as const;
 
-export const BUILTIN_MODULES_REGEX_STR = `^(?:node:)?(?:${builtinModules.join(
+export const BUILTIN_MODULES_REGEX_STR = `(^(?:node:)?(?:${builtinModules.join(
     '|',
-)})$`;
+)})$)|(^node:test$)`; // NOTE: 'test' is not included in `builtinModules` so explicitly check for "node:test"
```
IanVS, line 28: My only small suggestion is to add a link to the issue: https://github.com/nodejs/node/issues/42785.  Maybe someday it will be solved, and we will be able to remove this workaround.

IanVS (commented):
Looks awesome, thanks for the contribution.  I just had one minor request.

IanVS (approved):

IsaiahByDayah:
Sorry, left an extra white-space by accident

PR #152, opened 2024-02-28
Svelte import sort

secondfry:


src/index.ts
```diff
@@ -9,9 +9,12 @@ import {
     THIRD_PARTY_MODULES_SPECIAL_WORD,
 } from './constants';
 import { defaultPreprocessor } from './preprocessors/default';
+import { sveltePreprocessor } from './preprocessors/svelte';
 import { vuePreprocessor } from './preprocessors/vue';
 import type { PrettierOptions } from './types';
 
+const { parsers: svelteParsers } = require('prettier-plugin-svelte');
```
secondfry, line 16: `prettier-plugin-svelte` typings are misleading, they do export `parsers`.

secondfry:
It does not help with `<script lang="ts">`. I will fix it and update the PR.

package.json
```diff
@@ -20,7 +20,7 @@
         "prepare": "yarn run compile",
         "compile": "tsc",
         "preexample": "yarn run compile",
-        "test": "cross-env NODE_OPTIONS=--experimental-vm-modules vitest --run",
+        "test": "cross-env NODE_OPTIONS=\" $NODE_OPTIONS --experimental-vm-modules \" vitest --run",
```
secondfry, line 23: This is for VSCode Javascript Debug Terminal.

secondfry:
Apparently I was mistaken – I thought that plugin does not work due to the fact that it didn't sort imports in such scenario:
```
import A from '../A.svelte';

import './app.css';

import B from 'svelte';
```

Plugin does indeed work and splits imports into chunks and does not sort around side effect imports.
Thus this PR is unneeded. Closing it.

PR #153, opened 2024-03-11
chore: upgrade babel parser version

zxt-tzx:
hello, bumping up the babel parser version because I am using `prettier-plugin-sort-imports` in my codebase and using  latest `using` syntax resulted in a Prettier error

I poked around and found out it's because this syntax is is supported only since v7.22.0 of babel: see https://babeljs.io/blog/2023/05/26/7.22.0

thanks!

### DIFF
--- a/package.json
+++ b/package.json
@@ -55,7 +55,7 @@
     "dependencies": {
         "@babel/core": "^7.21.8",
         "@babel/generator": "^7.21.5",
-        "@babel/parser": "^7.21.8",
+        "@babel/parser": "^7.24.0",
         "@babel/traverse": "^7.23.2",
         "@babel/types": "^7.21.5",
         "semver": "^7.5.2"
--- a/yarn.lock
+++ b/yarn.lock
@@ -164,7 +164,7 @@
     "@babel/parser" "^7.22.15"
     "@babel/types" "^7.22.15"
 
-"@babel/traverse@^7.21.5", "@babel/traverse@^7.23.2":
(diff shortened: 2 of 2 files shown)


IanVS:
Hi, thanks for the contribution.  We don't use the `using` syntax and we don't pin our version of Babel like the trivago project does.  So if you need to use a version of babel that is more recent than you have now, you should be able to just update your lockfile (either by deleting it and your node_modules and re-installing or by using a  `pnpm.overrides`, or `resolutions`, or whatever your package manager requires.  

I'll work on releasing a version soon that updates babel, but in the meantime at least you shouldn't be blocked.

IanVS:
Ah, thanks for sharing the error.  It sounds like you need to add `explicitResourceManagement` to `importOrderParserPlugins` in the prettier config.  If that doesn't work, would you mind opening an issue and sharing your prettier config so I can help troubleshoot there?

zxt-tzx:
> Ah, thanks for sharing the error. It sounds like you need to add `explicitResourceManagement` to `importOrderParserPlugins` in the prettier config. If that doesn't work, would you mind opening an issue and sharing your prettier config so I can help troubleshoot there?

will do, thank you so much!

zxt-tzx:
oh wait, after updating all the related dependencies and adding `explicitResourceManagement` to `importOrderParserPlugins` in the prettier config, the error has gone away.

no need for anything else (overriding or installing `@babel/plugin-proposal-explicit-resource-management`). 

thank you so much! will close this now.

PR #154, opened 2024-03-11
Update Babel to 7.24

IanVS:
This updates babel in our `dependencies` to be sure that folks are getting the most up-to-date version that supports new syntax like `using` (closes https://github.com/IanVS/prettier-plugin-sort-imports/pull/153).

I also re-generated our `yarn.lock` because I was getting some type errors, and removed `sourceFile` in one spot because it's no longer needed and throws a TypeScript error.

### DIFF
--- a/package.json
+++ b/package.json
@@ -53,16 +53,16 @@
     ],
     "license": "Apache-2.0",
     "dependencies": {
-        "@babel/core": "^7.21.8",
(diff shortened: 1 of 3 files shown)


fbartho (approved):

IanVS:
Thanks for the reviews, @fbartho!

Issue #159, opened 2024-03-15
Feature Request: Targeting Only Subdirectories in a Monorepo


FrancesCoronel:
**Is your feature request related to a problem?**

I want to introduce a standardized import order at my company. After some research, this seems like the best plugin to do that as it does not re-order across side-effect imports and is more actively maintained than the Trivago plugin. 🎉 

After some discussion with our FE infra team, they're open to incrementally applying this standard to specific subdirectories until we eventually get to the root.

However, it doesn't look like targeting subdirectories is currently possible with this plugin.

**Describe the solution you'd like**

I want to use Prettier's built-in configuration overrides method to apply import sorting only to specific subdirectories.

https://prettier.io/docs/en/configuration.html#configuration-overrides

Here is an example of how I would update the root `.prettierrc` file to target certain directories.

Note that I introduced `"importOrder": null" as a way to not require sorting imports for all files except the ones defined in overrides.

```
{
	"printWidth": 100,
	"tabWidth": 4,
	"useTabs": true,
	"singleQuote": true,
	"importOrder": null,
	"overrides": [
		{
			"files": [
				"js/modern/components/WIP/feature/**/*.{ts,tsx,js,jsx}",
				"js/modern/redux/stores/feature-store/**/*.{ts,tsx,js,jsx}"
			],
			"options": {
				"importOrder": [
					"<THIRD_PARTY_MODULES>",
					"",
					"^react$",
					"",
					"^@feature/(?!.*\\.(less|css)$)(.*)$",
					"",
					"^@root/img/(.*)$",
					"",
					"^[./]",
					"",
					".*\\.(less|css)$"
				]
			}
		}
	]
}
```

**Describe alternatives you've considered**

The only alternative would be to format the entire monorepo simultaneously, which would be very risky given the size of the diff and because we are targeting imports which can lead to run time errors if incorrectly updated.

**Additional context**

Thank you for all your work on this! 💯

IanVS:
@FrancesCoronel thanks for the feature request!  I did a little bit of experimenting and digging around, and found that Prettier v4 (alpha) does not support overrides for plugin options at all right now (see https://github.com/prettier/prettier-cli/issues/9).  So if you're using that version, then unfortunately we're out of luck until that is resolved.

[...]

PR #161, opened 2024-03-15
Allow disabling plugin by setting importOrder to `[]`

IanVS:
Closes https://github.com/IanVS/prettier-plugin-sort-imports/issues/159

This changes up how we handle `importOrder: []` in the prettier config.  Previously, we would have applied our default config, such that builtins, node modules, and relative imports would be sorted.  Now, it effectively disables the plugin so that no sorting takes place.

This could be seen as a breaking change, but instead I see it as a bug fix.  It doesn't make sense that using an explicit empty array for the setting causes a fallback to be used.  

Additionally, this gives users more control over disabling the plugin and turning it on only for certain subfolders or types of files, as shown in the referenced issue and documented in the readme in this PR.


fbartho (approved):

IanVS:
Thanks for the reviews!

PR #172, opened 2024-06-06
fix: augment Prettier `Options` types with internal `PluginConfig`

### LINKED ISSUE #169: Type `RequiredOptions` in "prettier" module is not augmented with plugin options
**Your Environment**

-   **Prettier version**: 2.x.x
-   **node version**: v22.2.0
-   **package manager**: pnpm@9.1.2
-   **IDE**: VSCode (v1.89.1)

**Describe the bug**

TypeScript [module augmentation](https://www.typescriptlang.org/docs/handbook/declaration-merging.html#module-augmentation) with the `"prettier"` module is not supported, meaning using JSDoc types for the plugin options are not correctly denoted in IDEs (and presumably when the prettier config are typed-checked with TypeScript).

**To Reproduce**

1. Open an IDE with TypeScript support.
2. See what type is given for any of the plugin options (e.g. `importOrder`, `importOrderTypeScriptVersion`, `importOrderParserPlugins`).
3. See it is based on what is inferred (usually `never[]` or `string[]`) and does not have the corresponding JSDoc comments.

**Expected behavior**

As shown in the second screenshot, I expect the correct types for the options to be given, but also when using the `Config` type from `prettier` (first screenshot). It's a minor quibble, since `import("@ianvs/prettier-plugin-sort-imports").PrettierConfig` is also provided, but that is a bit long.

I believe the code that would make this possible something like:

```js
// @ianvs/prettier-plugin-sort-imports/types/index.d.ts
declare module 'prettier' {
  interface RequiredOptions extends PluginConfig {}
}
```

but when I attempted the trivial fix, I had some conflicts with the types in `src/types.ts`. I will take a look further, but I wanted to let you know beforehand.

image

image

```js
// .prettierrc.js
/** @type {import(prettier").Config} */
module.exports = {
  plugins: [require("@ianvs/prettier-plugin-sort-imports")],
  importOrder: [], // never[], but should be string[] | undefined
};
```

**Error log**

No errors.

-   [x] I'm willing to fix this bug 🥇

jeremy-code:
- Add TypeScript module augmentation for the `prettier` module in `types/index.d.ts` to add the `PluginConfig` interface to the `Options` interface to support proper typing of plugin options in IDEs and during type-checking
- Closes #169
- Compiles and passes tests

### DIFF
--- a/README.md
+++ b/README.md
@@ -128,7 +128,7 @@ Add your preferred settings in your prettier config file.
 ```ts
 // @ts-check
 
-/** @type {import("@ianvs/prettier-plugin-sort-imports").PrettierConfig} */
+/** @type {import("prettier").Config} */
 module.exports = {
     // Standard prettier options
     singleQuote: true,
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -72,3 +72,7 @@ export interface PluginConfig {
 }
 
 export type PrettierConfig = PluginConfig & Config;
+
+declare module 'prettier' {
+    interface Options extends PluginConfig {}
+}


IanVS:
Awesome, this looks great, thanks!  I tested it out in my project, and it worked well.  Would you mind updating the README along with this change, though, to suggest the use of `/** @type {import("prettier").Config} */` instead of `/** @type {import("@ianvs/prettier-plugin-sort-imports").PrettierConfig} */`?

jeremy-code:
Sounds good, made the requested change to [README.md](https://github.com/IanVS/prettier-plugin-sort-imports/pull/172/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5)

Issue #162, opened 2024-03-19
Formatting Vue SFC files error with @vue/compiler-sfc@2.7.16


shalldie:
**Your Environment**

-   **Prettier version**: 3.2.5
-   **node version**: 20.11.1
-   **package manager**: [npm@7, pnpm@6, yarn@2] pnpm@8.15.4
-   **IDE**: [VScode, Webstorm, CLI] vscode

**Describe the bug**

An error appear with `@vue/compiler-sfc@2.7.16`，but works well with `@vue/compiler-sfc@3.4.21`

**To Reproduce**

I made a respository which can reproduce this.
https://github.com/shalldie/sort-import-err-in-compiler-sfc-2.7.16

And here's the steps (also in README.md):

#### Case @vue/compiler-sfc@2.7.16

Now is `@vue/compiler-sfc@2.7.16`

1. pnpm i
2. format the file `src/App.vue`
3. it shows error:

```
TypeError: Cannot read properties of undefined (reading 'match')
```

#### Case @vue/compiler-sfc@3.4.21

1. Edit `package.json` to change the version to `@vue/compiler-sfc@3.4.21`
2. run `pnpm i`
3. format the file `src/App.vue`
4. It works well.

**Contribute to @ianvs/prettier-plugin-sort-imports**

-   [ ] I'm willing to fix this bug 🥇

shalldie:
> @shalldie I've released version 4.3.0 which fixes the bug you reported here. Please try it out and let me know if it's working correctly for you now. And thanks for the report with the reproduction, that was very helpful!

`4.3.0` does a good work !

PR #173, opened 2024-06-25
Fix vue 2.7 support

IanVS:
Fixes https://github.com/IanVS/prettier-plugin-sort-imports/issues/162

Older versions (not sure which ones exactly) of vue/compiler-sfc do not include a `loc` property on each block, but rather give the offsets directly on the block itself.  This accounts for those differences, and causes the reproduction provided in the issue to succeed.

### DIFF
--- a/src/preprocessors/vue.ts
+++ b/src/preprocessors/vue.ts
@@ -36,8 +36,8 @@ export function vuePreprocessor(code: string, options: PrettierOptions) {
             // [start, end)
 
             // @ts-expect-error Some vue versions have a `block.loc`, others have start and end directly on the block
-            let {start, end} = block;
-            if ("loc" in block) {
+            let { start, end } = block;
+            if ('loc' in block) {
                 start = block.loc.start.offset;
                 end = block.loc.end.offset;
             }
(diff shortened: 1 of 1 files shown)


src/preprocessors/vue.ts
```diff
@@ -48,8 +54,8 @@ export function vuePreprocessor(code: string, options: PrettierOptions) {
             console.warn(
                 '[@ianvs/prettier-plugin-sort-imports]: Could not process .vue file.  Please be sure that "@vue/compiler-sfc" is installed in your project.',
             );
-            throw err;
         }
+        throw err;
```
fbartho, line 58: Is it appropriate to always rethrow the error? In what cases do we not want to log the warning but still rethrow?

fbartho (approved):
:+1: This is more or less a rubber-stamp approval — I don’t use vue. I think your comment is sufficient documentation of the `@ts-expect-error` / edge case!

src/preprocessors/vue.ts
IanVS: The warning is an attempt to help guide the user towards a solution.  It detects when there is a `MODULE_NOT_FOUND`, and takes a guess that it's probably a missing dependency, so it gives that tip.  

I think it was just a mistake that it swallowed the error in the past, rather than re-throwing it.

PR #174, opened 2024-06-25
Add support for import attributes

IanVS:
There are three styles of import assertions/attributes that have been proposed over the years and made it to various stages of TC-39.  Here's a bit of [history](https://github.com/tc39/proposal-import-attributes#history): 

1) `import x from 'x' with type: "json"` - original proposal, made it to stage 2
2) `import x from 'x' assert { type: 'json' };` - import "assertion", got to stage 3, then found problems
3) `import x from 'x' with { type: 'json' };` - import "attribute", current proposal, stage 3

Babel only supports these with the use of a parser plugin, either `importAttributes` or `importAssertions`.  Also, babel/generator has to be told which style to generate.  Upstream has a PR, https://github.com/trivago/prettier-plugin-sort-imports/pull/273/files, which adds another user-level option to control this.  Instead, I've taken the stance here that we should only generate the latest "import attribute" style code.  So, not only does this PR add support for import attributes/assertions, this has the side-effect of converting deprecated import assertions to import attributes.  Maybe there's some reason someone would want to stick with the old style, but I can't think of one, and I personally would appreciate getting updated to the latest format.

### DIFF
--- a/src/utils/__tests__/get-code-from-ast.spec.ts
+++ b/src/utils/__tests__/get-code-from-ast.spec.ts
@@ -77,7 +77,9 @@ test('handles import attributes and assertions, converting to attributes when ne
     import g from 'g' with { type: 'json' };
 import c from 'c' assert { type: 'json' };
 `;
-    const importNodes = getImportNodes(code, { plugins: [['importAttributes', {deprecatedAssertSyntax: true}]] });
+    const importNodes = getImportNodes(code, {
+        plugins: [['importAttributes', { deprecatedAssertSyntax: true }]],
+    });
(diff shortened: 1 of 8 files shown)


fbartho (approved):
I’m relatively opposed to import-attributes in the first place*
— So I definitely endorse you only supporting the latest syntax!

<hr>
It feels like a runtime concern that I don’t see why should be in the import expression — and further allows for differing behavior in different files, since there’s not just 1 legal import statement for the same file

IanVS:
Yeah I don't have experience or opinions on them myself, but wanted to get ahead of any issue reports since it's already something that has come up in the upstream package.

PR #175, opened 2024-06-25
Improve compat with other prettier plugins

IanVS:
This is based on the approach in https://github.com/hosseinmd/prettier-plugin-jsdoc/tree/master, which allows other plugins some ability to work alongside this plugin (so long as this one is specified last, or used together with another plugin that interoperates.

I think there's the possibility of an infinite loop though depending on the order that plugins are specified, which I still need to look into in order to make sure we don't break existing users.

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -1,10 +1,12 @@
-// Based on node 16 config: https://github.com/tsconfig/bases/blob/main/bases/node16.json
+// Based on node 18 config: https://github.com/tsconfig/bases/blob/main/bases/node18.json
 {
     "$schema": "https://json.schemastore.org/tsconfig",
-    "display": "Node 16",
+    "display": "Node 18",
     "compilerOptions": {
-        "target": "es2021",
-        "module": "NodeNext",
+        "lib": ["es2023"],
+        "module": "node16",
+        "moduleResolution": "node16",
+        "target": "es2022",
         "allowJs": true,
         "outDir": "./lib",
         "rootDir": "./",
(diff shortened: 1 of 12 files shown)


tsconfig.json
```diff
@@ -1,10 +1,12 @@
...
     "$schema": "https://json.schemastore.org/tsconfig",
-    "display": "Node 16",
+    "display": "Node 18",
     "compilerOptions": {
-        "target": "es2021",
-        "module": "NodeNext",
+        "lib": ["es2023"],
+        "module": "node16",
+        "moduleResolution": "node16",
+        "target": "es2022",
```
IanVS, line 9: Changing from 2021 to 2022 could have an impact on the minimum version of node that we support, basically changing the minor version of node 16 that is required.  Whether we actually _use_ any of the features that aren't supported in early 16 versions, I'm not sure.  So maybe best to save this for a breaking version later on.

IanVS:
Confirmed, there's an infinite loop that occurs if this change is made and the plugin is used together with `prettier-plugin-jsdoc`.  Maybe there's a way around it, but I don't think so, at least not with this approach.  We'd need to do something like tailwindcss, with an [allowlist](https://github.com/tailwindlabs/prettier-plugin-tailwindcss/blob/f69f739537024cde95b85d0b9ea7aca0a50fc858/src/plugins.ts#L168) of compatible plugins, but that seems like too much work right now.

I'm going to close this out, and propose that we don't worry about it unless we hear from users that are having problems with other plugins.  And hopefully in the meantime Prettier will create a better API for plugins to work together in a pipeline.

PR #178, opened 2024-07-12
Internal: Switch from yarn to pnpm

IanVS:
It's faster, less buggy than yarn 1.x, and easier to work with than yarn 3/4.

One note: I did not try to migrate the lockfile, so this also updates all dependencies based on semver range in our package.json


fbartho (commented):
pnpm wasn’t safe to use with react-native when I last checked (>3 years ago). Is that fixed now?

Forgive a perhaps silly question, but if we do this, it implies a switch from yarn.lock to pnpm-lock.yaml — do package consumers end up seeing that file or no? My brain isn’t fully booted on this one.

IanVS:
I'm not sure about pnpm and react-native, but since this isn't a react-native project, I think it's fine.  The change will have no impact on consumers of the project, since whatever package manager _they_ are using will be used to install it.  So to answer your other question, no, the lock file is not exposed to end-users of a library / package.  It's only used by contributors of the project when installing our dependencies and working on the project.

fbartho (approved):

PR #184, opened 2024-09-27
Re-add `importOrderCaseSensitive` option for case-sensitive sorting

### LINKED ISSUE #151: Add importOrderCaseInsensitive option again
There's value in sorting uppercase first: constants, constructors and functions become sorted in that order.

Support for case sensitive sorting was dropped here #63

acnebs:
Implements a custom natural sort algorithm which allows for numeric natural sorting while also sorting all uppercase letters before all lowercase letters, which is the desired behavior here.

For example, when `importOrderCaseSensitive` is false (the default):

```js
import ExampleComponent from './ExampleComponent';
import ExamplesList from './ExamplesList';
import ExampleWidget from './ExampleWidget';
```

Compared with `"importOrderCaseSensitive": true`:

```js
import ExampleComponent from './ExampleComponent';
import ExampleWidget from './ExampleWidget';
import ExamplesList from './ExamplesList';
```

Closes #151

### DIFF
--- a/README.md
+++ b/README.md
@@ -407,6 +407,7 @@ used to order imports within each match group.
 For example, when false (or not specified):
 
 ```javascript
+import {CatComponent, catFilter, DogComponent, dogFilter} from './animals';
 import ExampleComponent from './ExampleComponent';
 import ExamplesList from './ExamplesList';
 import ExampleWidget from './ExampleWidget';
@@ -418,6 +419,7 @@ compared with `"importOrderCaseSensitive": true`:
 import ExampleComponent from './ExampleComponent';
 import ExampleWidget from './ExampleWidget';
 import ExamplesList from './ExamplesList';
+import {CatComponent, DogComponent, catFilter, dogFilter} from './animals';
 ```
 
 ### Prevent imports from being sorted
--- a/README.md
+++ b/README.md
@@ -34,6 +34,7 @@ This project is based on [@trivago/prettier-plugin-sort-imports](https://github.
       - [7. Enable/disable plugin or use different order in certain folders or files](#7-enabledisable-plugin-or-use-different-order-in-certain-folders-or-files)
     - [`importOrderTypeScriptVersion`](#importordertypescriptversion)
     - [`importOrderParserPlugins`](#importorderparserplugins)
+    - [`importOrderCaseSensitive`](#importordercasesensitive)
   - [Prevent imports from being sorted](#prevent-imports-from-being-sorted)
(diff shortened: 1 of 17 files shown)


tests/ImportOrderCaseSensitive/__snapshots__/ppsi.spec.ts.snap
```diff
@@ -0,0 +1,20 @@
...
+import {cat, Dog} from "C";
+import B from "B";
+import A from "A";
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+import a from "a";
+import A from "A";
+import b from "b";
+import B from "B";
+import c from "c";
+import { cat, Dog } from "C";
```
IanVS, line 18: I added an end-to-end test, and it looks like this is not behaving as expected.  Would you mind looking into it to see what's not quite wired up yet?

IanVS (changes requested):
Thanks for the PR!  The sorting function looks solid, but it seems like maybe the option is not wired completely through, judging by an end-to-end test that I added.  Would you be willing to look into that to see where it might be broken?

acnebs:
Yep will investigate.

acnebs:
That should fix it @IanVS

acnebs:
Would it be possible to merge this in now that it is passing all checks?

README.md
```diff
@@ -393,6 +395,31 @@ with options as a JSON string of the plugin array:
...
+```
+
+compared with `"importOrderCaseSensitive": true`:
+
+```javascript
+import ExampleComponent from './ExampleComponent';
+import ExampleWidget from './ExampleWidget';
+import ExamplesList from './ExamplesList';
+```
+
```
IanVS, line 422: Since this will also sort ImportSpecifiers (`import {cat, Dog} from "C";` -> `import {Dog, cat} from "C";`), I think we should probably also include that in the readme here.

IanVS (commented):
Just one more small ask for the readme, then I think this will be ready to merge.  Sorry for the delay, I was at a conference much of last week.

acnebs:
Readme updated 👍

IanVS (approved):

IanVS:
Thanks so much for your work on this!  I'll do my best to get a new minor version released soon, but there's a thing or two that I want to squeak in with it as well.

PR #198, opened 2025-01-12
fix: catch babel parse errors

### LINKED ISSUE #197: Plugin throws less helpful errors when code has a SyntaxError
**Your Environment**

-   **Prettier version**: 3.4.2
-   **node version**: 22.12.0
-   **package manager**: pnpm@9.15.2
-   **IDE**: VSCode

**Describe the bug**

I've noticed that when a file has legitimate syntax errors, that this plugin will throw before prettier is able to throw.  I suspect it's just uncaught `babel/traverse` errors, and since the plugin is a preprocessor, it crashes the prettier run before prettier is able to throw.  That would be ok, if the errors from this plugin / babel were at least as helpful as the errors from prettier.  But what I've found is that they're aren't.  I'd like to propose that we wrap the babel parsing in a try catch and just write the errors to console, rather than throwing, so that we give prettier a chance to throw more helpful errors.

**To Reproduce**

Repro repo: https://github.com/michaelfaith/ianvs-prettier-plugin-sort-imports-bug

screenshot of the repro

(first execution is running prettier without the plugin, and the second run is with the plugin)

-   [x] I'm willing to fix this bug 🥇

michaelfaith:
This change updates the plugin to catch babel parsing errors, so that we can give prettier a chance to throw.  We're still writing an error to console, so that in the event that prettier doesn't end up throwing, we're still indicating that the import sorting failed.

screenshot of new behavior

Closes #197

### DIFF
--- a/src/preprocessors/preprocessor.ts
+++ b/src/preprocessors/preprocessor.ts
@@ -26,7 +26,9 @@ export function preprocessor(code: string, options: PrettierOptions): string {
     try {
         ast = babelParser(code, parserOptions);
     } catch (_) {
-        console.error(' [error] [prettier-plugin-sort-imports]: import sorting aborted due to babel parsing error.');
+        console.error(
+            ' [error] [prettier-plugin-sort-imports]: import sorting aborted due to babel parsing error.',
+        );
         return code;
     }
 
--- a/src/preprocessors/preprocessor.ts
+++ b/src/preprocessors/preprocessor.ts
@@ -22,7 +22,13 @@ export function preprocessor(code: string, options: PrettierOptions): string {
         parserOptions.allowReturnOutsideFunction = true;
     }
 
-    const ast = babelParser(code, parserOptions);
+    let ast;
+    try {
+        ast = babelParser(code, parserOptions);
+    } catch (_) {
+        console.error(' [error] [prettier-plugin-sort-imports]: import sorting aborted due to babel parsing error.');
+        return code;
+    }
 
     const directives = ast.program.directives;
     const interpreter = ast.program.interpreter;
--- a/src/preprocessors/preprocessor.ts
+++ b/src/preprocessors/preprocessor.ts
@@ -22,7 +22,7 @@ export function preprocessor(code: string, options: PrettierOptions): string {
         parserOptions.allowReturnOutsideFunction = true;
     }
 
-    let ast;
+    let ast: ReturnType<typeof babelParser>;
     try {
         ast = babelParser(code, parserOptions);
     } catch (_) {


src/preprocessors/preprocessor.ts
```diff
@@ -22,7 +32,13 @@ export function preprocessor(code: string, options: PrettierOptions): string {
         parserOptions.allowReturnOutsideFunction = true;
     }
 
-    const ast = babelParser(code, parserOptions);
+    let ast;
+    try {
+        ast = babelParser(code, parserOptions);
```
michaelfaith, line 37: I said in the original issue that it was babel.traverse that was throwing, but I mispoke; it was actually the parse that happens first.  So, I only wrapped the parse itself.

src/preprocessors/preprocessor.ts
```diff
@@ -7,6 +7,16 @@ import { getCodeFromAst } from '../utils/get-code-from-ast';
...
 import { examineAndNormalizePluginOptions } from '../utils/normalize-plugin-options';
 
+const colors = {
+    red: '\x1b[31m',
+    yellow: '\x1b[33m'
+} as const;
+
+function colorText(text: string, color: keyof typeof colors): string {
+    const resetColor = '\x1b[0m';
+    return `${colors[color]}${text}${resetColor}`
```
fbartho, line 17: We need to detect if the terminal/process/prettier has colors enabled before posting with colors, so that if they're piping this, we don't bake in these terminal codes for an environment that doesn't support it.

fbartho (changes requested):
Looks good to me, other than the change needed for colors!

Thanks!

src/preprocessors/preprocessor.ts
michaelfaith: Went ahead and removed the colors.  Doesn't seem worth adding another library (or adding all the required code locally) for that.

src/preprocessors/preprocessor.ts
```diff
@@ -22,7 +22,13 @@ export function preprocessor(code: string, options: PrettierOptions): string {
         parserOptions.allowReturnOutsideFunction = true;
     }
 
-    const ast = babelParser(code, parserOptions);
+    let ast;
```
fbartho, line 25:
```suggestion
    let ast: ReturnType<typeof babelParser>;
```

fbartho (approved):
@IanVS will have to do the next steps, but this looks good to me!

thanks again!

IanVS:
Thanks a ton, @michaelfaith, this will be very helpful.

IanVS (approved):

PR #210, opened 2025-05-26
Update peerDeps to allow prettier v4.

cpojer:
I've been using the v4 alphas of prettier for a *very long time*, and they have always worked just fine with this plugin.

### DIFF
--- a/package.json
+++ b/package.json
@@ -77,7 +77,7 @@
     },
     "peerDependencies": {
         "@vue/compiler-sfc": "2.7.x || 3.x",
-        "prettier": "2 || 3"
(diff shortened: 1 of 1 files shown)


IanVS:
Thanks Christoph, I also use v4 alpha of prettier, I should have done this long ago.

Are you sure that `4` will pick up the alpha versions?  I don't 100% trust the semver calculator at https://semver.npmjs.com/, but it indicates that we'd need to use something like `^4.0.0-0`.  I know peer deps are sometimes treated differently from standard deps, but maybe that would be safer?

cpojer:
Honestly, I don't know 🥲

PR #211, opened 2025-05-27
[Internal] Update prettier dev dependency to 4.0 alpha

IanVS:
Updates the version in our development dependencies to 4.0.0-alpha.12, which runs more quickly than the 3.x version (see https://prettier.io/blog/2023/11/30/cli-deep-dive).

As an added bonus, prettier is now working again in vscode when I work on the project (I think we had a bad config previously).

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -16,10 +16,10 @@ assignees:
 
 **Your Environment**
 
--   **Prettier version**: 2.x.x
--   **node version**:
--   **package manager**: [npm@7, pnpm@6, yarn@2]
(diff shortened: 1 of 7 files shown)


fbartho (approved):
No material objections to the content of this PR. (I haven’t tested — let me know if you need that)

<hr>

I’m mildly concerned how long the `4.0.0-alpha` line has existed. Do we have a sense on when it’s going to be tagged `latest`?

It does look like they’ve added a new alpha each time they’ve added a new `latest`, so maybe it’s not likely to get desynched which changes that ship in `3.x` but not in `4.0.0-alpha`.

IanVS:
I don't have any insight into the eventual release of 4 as stable.  But I did check like you did to see that it's keeping up-to-date, and so far it has been.  :+1:

Issue #220, opened 2025-06-25
Incompatibility with the new experimental `@prettier/plugin-oxc` plugin


ThenTech:
**Your Environment**

- **Prettier version**: 3.6.1
- **node version**: 22.13.1
- **package manager**: pnpm@10.12.3
- **IDE**: VSCode

**Describe the bug**

Using both this and the new `@prettier/plugin-oxc`, causes the import sorting not to happen. Other formatting does occur as expected. Removing the `@prettier/plugin-oxc` plugin also restores sorting imports.

Perhaps the new OXC based plugin is skipping other plugins, or does not parse/provide the required import statements to sort imports in the first place?

**To Reproduce**

1. Have both this plugin and `@prettier/plugin-oxc` in the Prettier config (`.prettierrc.json` in my case):
```jsonc
// ...
"plugins": [
    "@ianvs/prettier-plugin-sort-imports",
    "@prettier/plugin-oxc"
],
// ...
```
2. Format a file with unsorted imports
3. File is formatted, but imports are untouched

**Expected behavior**

Both Prettier formatting and import sorting should happen.

**Error log**

No errors occur in the output log for Prettier in VSCode.

**Contribute to @ianvs/prettier-plugin-sort-imports**

- [ ] I'm willing to fix this bug 🥇

PR #221, opened 2025-07-06
feat: add plugin-oxc support

Marsup:
Fixes https://github.com/IanVS/prettier-plugin-sort-imports/issues/220.

I copied the tests for babel/babel-ts, hopefully this is acceptable.

The hard inclusion of plugin-oxc is a bit dirty but I'm not sure there's any other way.


package.json
```diff
@@ -57,6 +57,7 @@
         "@babel/parser": "^7.26.2",
         "@babel/traverse": "^7.25.9",
         "@babel/types": "^7.26.0",
+        "@prettier/plugin-oxc": "^0.0.4",
```
fbartho, line 60: Is there some way to make this an optional dependency?

Feels like we should only integrate this plugin *if* the user already has this plugin installed.

fbartho (changes requested):
No philosophical objection to this PR.

Just think the dependency should be optional if at all possible!

Thanks @Marsup!

package.json
fbartho: Maybe a function with an inline require? Only should be executed if the processor gets triggered. I'm fine if that has a `// ts-ignore` + maybe a try/catch to log a warning the user if they forgot to install a plugin, but prettier is trying to use it?

Marsup: Let me know if it's OK like that

package.json
```diff
@@ -71,6 +72,9 @@
         "typescript": "5.2.2",
         "vitest": "^0.34.5"
     },
+    "optionalDependencies": {
```
IanVS, line 75: This isn't quite what we want.  `optionalDependencies` will always be attempted to be installed, and the only difference is that `npm install` will not fail if the package fails to install (due to being built for a different computer architecture, for instance).  

Instead, we can move this to `peerDependencies`, and add an `optional: true` in `peerDependenciesMeta`, similar to what we have for `@vue/compiler-sfc`.

Marsup:
Oh sorry, didn't expect a change in the lockfile due to this, done.

IanVS:
I took a look into the test failures, and the error being caught in `getOxcPlugin` is:

```
Error [ERR_REQUIRE_ESM]: require() of ES Module /Users/ianvs/code/ianvs/prettier-plugin-sort-imports/node_modules/.pnpm/@prettier+plugin-oxc@0.0.4/node_modules/@prettier/plugin-oxc/index.mjs not supported.
Instead change the require of /Users/ianvs/code/ianvs/prettier-plugin-sort-imports/node_modules/.pnpm/@prettier+plugin-oxc@0.0.4/node_modules/@prettier/plugin-oxc/index.mjs to a dynamic import() which is available in all CommonJS modules.
```

We might be able to use a dynamic import() and typescript should compile down to cjs for the production bundle.  I'll try that out.

IanVS:
Right.  This `ERR_REQUIRE_ESM ` is really only an issue in tests, since like you said we compile and ship CJS.  Upgrading to a more recent version of Node.js allows us to `require(esm)`, but there are a few other changes needed as well to support that.  Mind if I push a couple commits to this branch?

IanVS:
I did notice in my testing that `'@prettier/plugin-oxc'` needs to come before `'@ianvs/prettier-plugin-sort-imports'` in the `plugins` config array, so I've added a note to the README about it.  

Thanks again for your work on this, @Marsup!

@fbartho, what do you think, does this look good to merge?

fbartho:
Sorry I didn't reply yesterday. Got interrupted by work and forgot to submit. No objections now that it's optional!

IanVS:
Cool, thanks for taking a look.  I'm going to try to get a release out soon.  👍

PR #227, opened 2025-08-01
add support for ember template tags

### LINKED ISSUE #225: Support for glimmer javascript + typescript
I currently use `@ianvs/prettier-plugin-sort-imports` to sort imports in js + ts files in a project using the ember framework. They are starting to recommend a new component authoring format (single file components with gts + gjs extensions) which does not appear to work out of the box with this prettier plugin.

Would it be possible to add support for it?

I would like to be able to sort imports in GTS + GJS files with this plugin.

I have considered using eslint plugins which offer similar functionality, but I would like to stay on one dependency (i.e. this prettier plugin) for import sorting if I can.

**Additional context**
In order to get prettier working for the GTS + GJS components, you need this plugin: https://github.com/ember-tooling/prettier-plugin-ember-template-tag. Is this an issue? I know that there are some issues related to using multiple plugins on one file, but I'm not all that familiar with prettier under the hood to know whether this is do-able at this point or not.

I'm open to trying to submit a patch for this, but I want to sanity check that this is something that would be accepted and that is possible before diving into it too much.

geneukum:
Closes #225.

Let's add support for ember's single-file component format (gts + gjs). This is largely just a port of https://github.com/trivago/prettier-plugin-sort-imports/pull/377, so all credit really goes to the folks involved with that one.

That said, I'm still happy to make whatever changes needed to help this land here. Thanks again for the great plugin / considering this addition!


src/preprocessors/ember.ts
```diff
@@ -0,0 +1,63 @@
...
+    function ingest(collection?: {
+        [key: string]: readonly {
+            readonly start: number;
+            readonly end: number;
+        }[];
+    }) {
+        if (!collection) return;
+
+        for (let [, info] of Object.entries({ ...collection })) {
+            for (let pos of info) {
```
IanVS, line 37: Nit: I prefer using `const` for variables that are not intended to be changed, rather than `let`.

README.md
```diff
@@ -458,15 +458,16 @@ Having some trouble or an issue? You can check [FAQ / Troubleshooting section](.
...
+| JS with ES Modules     | ✅ Everything   | -                                                                                           |
+| NodeJS with ES Modules | ✅ Everything   | -                                                                                           |
+| React                  | ✅ Everything   | -                                                                                           |
+| Svelte                 | ✅ Everything   | -                                                                                           |
+| Angular                | ✅ Everything   | Supported through `importOrderParserPlugins` API                                            |
+| Vue                    | ✅ Everything   | SFCs only, peer dependency `@vue/compiler-sfc` is required                                  |
+| Astro                  | 🧪 Experimental | Some Astro syntax may cause trouble, please open an issue                                   |
+| Ember                  | 🧪 Experimental | peer dependencies `prettier-plugin-ember-template-tag` and `parse-imports-exports` required |
```
IanVS, line 470: I appreciate being conservative here and not requiring `parse-imports-exports` for non-Ember projects, but on the other hand I also kind of hate peer dependencies and this feels like an odd one to ask users to install manually.  I'm all for keeping the project size small, but especially since it sounds like we might be able to use this utility library for other syntaxes as well, how about we just add it to our dependencies?

tests/Ember/ppsi.spec.mjs
IanVS, line 1: Let's change this filename to `ppsi.spec.ts` so that it's not included in the snapshot.

tests/Ember/sfc.gts
```diff
@@ -0,0 +1,36 @@
...
+import sameLevelRelativePath from './sameLevelRelativePath';
+import thirdParty from 'third-party';
+import oneLevelRelativePath from '../oneLevelRelativePath';
+import otherthing from '@core/otherthing';
+import abc from '@core/abc';
+import twoLevelRelativePath from '../../twoLevelRelativePath';
+import component from '@ui/hello';
+import fourLevelRelativePath from '../../../../fourLevelRelativePath';
+import * as a from 'a';
+import type RouterService from '@ember/routing/router-service';
```
IanVS, line 13: let's add a few more styles like `import type {T} from...` and `import a, type {T}`, etc.  Just for good measure.

tests/Ember/sfc.gjs
```diff
@@ -0,0 +1,28 @@
+// I am top level comment in this file.
+import z from 'z';
+import threeLevelRelativePath from '../../../threeLevelRelativePath';
+import sameLevelRelativePath from './sameLevelRelativePath';
+import thirdParty from 'third-party';
+import oneLevelRelativePath from '../oneLevelRelativePath';
+import otherthing from '@core/otherthing';
+import abc from '@core/abc';
+import twoLevelRelativePath from '../../twoLevelRelativePath';
+import component from '@ui/hello';
```
IanVS, line 10: Can we add a namespace import somewhere in here?  `import * as ....`

IanVS (commented):
This seems relatively clean and straightforward.  I had a few, mostly minor, comments, but I think this looks good to move forward.  I'm still not 100% certain on the use of `parse-imports-exports` and why that is required.  I suppose it is because `@babel/traverse` cannot process `.gjs/.gts` files to extract import statements in the same way that we do for other files, is that your understanding as well?

@fbartho do you have any thoughts on this PR?

src/preprocessors/ember.ts
```diff
@@ -0,0 +1,63 @@
...
+        }
+    }
+
+    ingest(importsExports.namedImports);
+    ingest(importsExports.namespaceImports);
+    ingest(importsExports.typeNamedImports);
+    ingest(importsExports.typeNamespaceImports);
+
+    let output = preprocessor(justImports, options);
+    let result = output + code;
```
geneukum, line 56: Just noticed - this behaves a bit differently from the currently supported languages.

It looks like those sort imports relative to existing ones, where this just appends them all at the top of the file.

I think this is probably problematic if, for example, there's a license header at the top. This would move the imports above it, which is probably not what folks would want. 

Let's see if we can fix this.

fbartho: >It looks like those sort imports relative to existing ones, where this just appends them all at the top of the file.
>
>I think this is probably problematic if, for example, there's a license header at the top. This would move the imports above it, which is probably not what folks would want.

Full expected behavior documented in our README https://github.com/IanVS/prettier-plugin-sort-imports?tab=readme-ov-file#comments (included here for convenience):

- If you have one or more comments at the top of the file, we will keep them at the top.
- Comments on lines after the final import statement will not be moved. (Runtime-code between imports will be moved below all the imports).
- In general, if you place a comment on the same line as an Import Declaration or *Specifier, we will keep it attached to that same specifier if it moves around.
- Other comments are preserved, and are generally considered "leading" comments for the subsequent Import Declaration or *Specifier.

I would indeed expect the above behavior to be followed for Ember files as well, because as you said -- different behavior "is probably not what folks want".

README.md
IanVS: >Aside: We're might have a bug around that (when merging import nodes), but mostly we should be fine with support for those and/or I'm not sure there's a current use-case for assertions where merging nodes is relevant!

Ah, good catch on them not supporting import assertions, though that format is now deprecated syntax and "import attributes" are the standard: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import/with.

What kind of bug do you think we have with assertions / attributes?  Maybe you can open an issue?

fbartho: Nice catch on that being the deprecated concept. I vaguely remember hearing that, but it hasn’t been top of nice.

In the hypothetical merger scenario of something using `with` import-attributes, I think we’re arbitrarily picking one of the nodes to merge the other into (presumably the earliest node becomes the parent).

I’m not sure what we should do if they have import attributes:
1. Prevent the merge if they use attributes
1. Abort the merge if they don’t have _identical_ attributes
1. Abort the merge if they have _incompatible_ attributes.
1. Merge the attributes

I only know how to implement 1 & 2. No idea how to implement 3 & 4.

I’m kind of of the mindset of either doing nothing until somebody reports an issue (I have no idea who actually uses this feature, and how), or implementing one of 1 or 2.

Further question: Is it legal to merge something without attributes into an expression that has some attributes? — If that’s not legal, then we likely have a bug.

fbartho: Filed a placeholder ticket here: https://github.com/IanVS/prettier-plugin-sort-imports/issues/229

IanVS: @fbartho by the way, why do you say "parse-import-exports doesn't support Import Assertions"?  I played around with it a bit in the trivago repo (since they merged a very similar PR), and both assertions and attributes seem to work fine.

Issue #228, opened 2025-08-03
Should we integrate this upstream?


RobbieTheWagner:
I'm curious, was there a reason this was a fork and not contributed back upstream? It might be nice to have some of the changes in the original repo.

IanVS:
The original reason was that the upstream project sorted side-effect-only imports along with all the others, which caused problems like changing the resulting styles when importing multiple CSS files.  They did not want to change that behavior, so a few of us created this fork as a safer option.  Since then, we've added quite a few features and simplified the options, being a bit more opinionated (following the core philosophy of prettier).  As a fork, we are also able to react more quickly to user requests and release more frequently than upstream tends to.  

When you say it might be nice to have changes in the original repo, I'm not sure if you mean it would be nice to have our changes upstream, or pull in changes from upstream into this one.  If you know of any features in the Trivago repo which you think we should consider adding here, I'd be happy to consider them.  This project has more than 3.5 million downloads a month, and I'm committed to keeping it running smoothly.  If the upstream project wants to add our features or pull in our code they're welcome to, but I don't plan to open any such PRs myself.  I think we have somewhat different philosophies, and I don't see the harm in giving users a few different plugin options.

fbartho:
In support of @IanVS, I agree with everything he said.

As a general comment, I would originally have preferred to just use a plugin without needing to contribute at all (zero labor = best! -- "It just works!"™).

My perception was that Trivago was stagnant, and overly complex, and significant refactor work was needed to clean up the codebase. This complicated my attempts at contributing over there. + (There were important Pull Requests that are still not merged / officially reviewed 3+ years later)

I'm happy here, and not looking to switch, but I certainly wouldn't be offended if they merged this code back into trivago. Getting that done would be a socio-political challenge that I'm not particularly interested in tackling.

PR #231, opened 2025-08-04
add support for ember template tags

### LINKED ISSUE #225: Support for glimmer javascript + typescript

IanVS:
Closes #225.

Adds support for ember's single-file component format (gts + gjs).  

This iterates upon https://github.com/IanVS/prettier-plugin-sort-imports/pull/227, with significant changes.  Now, we will preprocess the source code in a similar way to `prettier-plugin-ember-template-tag`, replacing `<template>` tags (using https://github.com/embroider-build/content-tag) with placeholder comments of the correct length.  Then we'll parse that code, extract and sort import nodes, and substitute the new nodes into the _original_ code string. 

~This does end up adding `content-tag` for all users, but it's a very small utility library so it shouldn't hurt anything.~ It's now an optional peer dependency, and must be installed manually for users of Ember template tags.

### DIFF
--- a/README.md
+++ b/README.md
@@ -467,7 +467,7 @@ Having some trouble or an issue? You can check [FAQ / Troubleshooting section](.
 | Angular                | ✅ Everything   | Supported through `importOrderParserPlugins` API                                            |
 | Vue                    | ✅ Everything   | SFCs only, peer dependency `@vue/compiler-sfc` is required                                  |
 | Astro                  | 🧪 Experimental | Some Astro syntax may cause trouble, please open an issue                                   |
-| Ember template tags    | 🧪 Experimental | peer dependency `prettier-plugin-ember-template-tag` required                               |
+| Ember template tags    | 🧪 Experimental | peer dependencies `prettier-plugin-ember-template-tag` and `content-tag` required           |
 
 ## Contribution
 
--- a/package.json
+++ b/package.json
@@ -57,7 +57,6 @@
         "@babel/parser": "^7.26.2",
         "@babel/traverse": "^7.25.9",
         "@babel/types": "^7.26.0",
-        "content-tag": "^4.0.0",
         "semver": "^7.5.2"
     },
     "devDependencies": {
@@ -67,6 +66,7 @@
         "@types/node": "^24.0.12",
         "@types/semver": "^7.7.0",
         "@vue/compiler-sfc": "3.5.17",
+        "content-tag": "^4.0.0",
         "cross-env": "^7.0.3",
         "prettier": "^3.6.2",
         "prettier-plugin-astro": "^0.14.0",
@@ -77,6 +77,7 @@
     "peerDependencies": {
         "@prettier/plugin-oxc": "^0.0.4",
         "@vue/compiler-sfc": "2.7.x || 3.x",
+        "content-tag": "^4.0.0",
         "prettier": "2 || 3 || ^4.0.0-0",
         "prettier-plugin-ember-template-tag": "^2.1.0"
     },
@@ -85,6 +86,9 @@
             "optional": true
         },
         "@vue/compiler-sfc": {
+            "optional": true
+        },
+        "content-tag": {
             "optional": true
         },
         "prettier-plugin-ember-template-tag": {
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -20,9 +20,6 @@ importers:
       '@babel/types':
         specifier: ^7.26.0
         version: 7.28.0
-      content-tag:
-        specifier: ^4.0.0
-        version: 4.0.0
       semver:
         specifier: ^7.5.2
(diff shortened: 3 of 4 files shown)


RobbieTheWagner:
I'm curious why there is a fork though? Seems like if there were improvements, they could be contributed upstream?

IanVS:
@RobbieTheWagner did you read my reply to your issue? https://github.com/IanVS/prettier-plugin-sort-imports/issues/228.  I  answered that question there.

IanVS:
@fbartho thanks for the review!  

Part of the issue here is that I don't really trust peer dependencies.  I've been bitten in the butt by them on several other projects, and they just turn into a maintenance nightmare.  But, after you raised your concern, I did try testing it out here.  I created a new ember project, added the ember template tag prettier plugin and this plugin, used `pnpm patch` to bring in the changes from this PR (after modifying it to make `content-tag` a peer-dep), created a couple of simple ember components with template tags, and tried to run prettier on them.  I got parse errors because ember itself includes an older version of `content-tag`, and that's what was being chosen unless I explicitly installed `content-tag` with the correct version as a direct dependency of the project.  But pnpm didn't warn me that I needed to do this, I think because I had marked it as an optional peer dependency.  

But I agree that when our package is only ~6mb currently, adding a 2.4mb dependency is quite a hit, especially since only a tiny fraction of users will need it.  So, I've made it a peer dependency and added a runtime check that the version we're seeing is compatible with what we need.

As for your other comment, I've never considered the preprocessor itself as part of our public API. It's not documented anywhere, and I can't think of any reason anyone would need to use it directly.  The next time we make a major release, we can add a `exports` field to our package.json to ensure that nobody is relying on any internal parts of the plugin, but for now I think we're safe to release this in a minor.  Thanks for bringing that up, though!

fbartho (commented):
I see you’ve configured `content-tag` to be optional, but I haven’t tested yet to see if the base library works without `content-tag` installed.

I’m not sure I’ll have time to test that until Sunday, but if that’s something you’re confident is working, then I have no remaining concerns about this PR.

If you’re willing to wait till then, then I’m happy to try to verify it by then.

IanVS:
@fbartho Just a note:  I just learned about `import defer`.  Seems like it could be useful for cases like this where we don't want to try to import a dependency unless some condition is met.  https://devblogs.microsoft.com/typescript/announcing-typescript-5-9/#support-for-import-defer

fbartho:
Today I Learned! -- Also I equally got disappointed that it can't be used to avoid importing an optional resource.

RobbieTheWagner:
This does not work in my project. I am getting errors like this:

```
app/templates/welcome/auto-start.gts
[error] app/templates/welcome/auto-start.gts: TypeError: Cannot read properties of undefined (reading 'ember-template-tag')
[error]     at get ember-template-tag [as ember-template-tag] (/Users/rwwagner90/shipshape/swach/node_modules/@ianvs/prettier-plugin-sort-imports/lib/src/index.js:72:35)
[error]     at initParser (file:///Users/rwwagner90/shipshape/swach/node_modules/prettier/index.mjs:19898:52)
[error]     at normalizeFormatOptions (file:///Users/rwwagner90/shipshape/swach/node_modules/prettier/index.mjs:19947:24)
[error]     at formatWithCursor (file:///Users/rwwagner90/shipshape/swach/node_modules/prettier/index.mjs:20727:11)
[error]     at Module.<anonymous> (file:///Users/rwwagner90/shipshape/swach/node_modules/prettier/index.mjs:22177:12)
[error]     at async formatFiles (file:///Users/rwwagner90/shipshape/swach/node_modules/prettier/internal/cli.mjs:4213:18)
[error]     at async main (file:///Users/rwwagner90/shipshape/swach/node_modules/prettier/internal/cli.mjs:4865:5)
[error]     at async Module.run (file:///Users/rwwagner90/shipshape/swach/node_modules/prettier/internal/cli.mjs:4811:5)
```

Seems like it cannot find `prettier-plugin-ember-template-tag`? I do have it installed though.

PR #234, opened 2025-08-13
fix: conditionally register ember and oxc parsers when dependencies available

### LINKED ISSUE #233: Error: prettier-plugin-ember-template-tag could not be loaded.  Is it installed?
**Your Environment**

- **Prettier version**: 3.6.2
- **node version**: 22.18.0
- **package manager**: npm@10.9.3
- **IDE**: [CLI]

**Describe the bug**

There seems to be an incompatibility with the newly added ember template functionality when using this plugin with prettier-plugin-toml

**To Reproduce**

```sh
git clone https://github.com/jahands/prettier-sort-imports-plugin-repro.git
cd prettier-sort-imports-plugin-repro
npm i
npm run format
```

**Expected behavior**

expect it not to error like it did in earlier versions of this plugin (before ember template support was added)

See minimal repro here: https://github.com/jahands/prettier-sort-imports-plugin-repro

**Error log**

```
$ npm run format

> prettier-sort-imports-plugin-repro@1.0.0 format
> prettier --write .

.prettierrc.cjs 17ms (unchanged)
example.toml
[error] example.toml: Error: Error: prettier-plugin-ember-template-tag could not be loaded.  Is it installed?
[error]     at M.format (/Users/jh/src/prettier-sort-imports-plugin-repro/node_modules/@taplo/lib/dist/index.js:2:35638065)
[error]     at format (file:///Users/jh/src/prettier-sort-imports-plugin-repro/node_modules/prettier-plugin-toml/lib/index.js:10:21)
[error]     at async parse5 (file:///Users/jh/src/prettier-sort-imports-plugin-repro/node_modules/prettier/index.mjs:16731:11)
[error]     at async coreFormat (file:///Users/jh/src/prettier-sort-imports-plugin-repro/node_modules/prettier/index.mjs:17287:25)
[error]     at async formatWithCursor (file:///Users/jh/src/prettier-sort-imports-plugin-repro/node_modules/prettier/index.mjs:17504:14)
[error]     at async formatFiles (file:///Users/jh/src/prettier-sort-imports-plugin-repro/node_modules/prettier/internal/legacy-cli.mjs:4279:18)
[error]     at async main (file:///Users/jh/src/prettier-sort-imports-plugin-repro/node_modules/prettier/internal/legacy-cli.mjs:4698:5)
[error]     at async Module.run (file:///Users/jh/src/prettier-sort-imports-plugin-repro/node_modules/prettier/internal/legacy-cli.mjs:4641:5)
package-lock.json 15ms (unchanged)
package.json 1ms (unchanged)
README.md 13ms (unchanged)
```

- [x] I'm willing to fix this bug 🥇

jahands:
Prevents errors when prettier-plugin-ember-template-tag or @prettier/plugin-oxc are not installed. The previous getter-based lazy loading caused errors during parser enumeration by other plugins (like prettier-plugin-toml), breaking compatibility. Now parsers are only registered when dependencies exist.

resolves #233

## Testing

I tested by removing the spread operators and confirmed relevant tests fail.
I also tested with my [repro](https://github.com/jahands/prettier-sort-imports-plugin-repro#) posted in #223 and confirmed it no longer errors

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -67,6 +67,15 @@ const getEmberPlugin = () => {
     }
 };
 
+const isEmberPluginAvailable = () => {
+    try {
+        getEmberPlugin();
+        return true;
+    } catch {
+        return false;
+    }
+};
+
 const getOxcPlugin = () => {
     try {
         const oxcPlugin = require('@prettier/plugin-oxc');
@@ -79,6 +88,15 @@ const getOxcPlugin = () => {
     }
 };
 
+const isOxcPluginAvailable = () => {
+    try {
+        getOxcPlugin();
+        return true;
+    } catch {
+        return false;
+    }
+};
+
 export const parsers = {
     babel: {
         ...babelParsers.babel,
@@ -88,34 +106,26 @@ export const parsers = {
         ...babelParsers['babel-ts'],
         preprocess: defaultPreprocessor,
     },
-    get 'ember-template-tag'() {
-        const emberPlugin = getEmberPlugin();
-
-        return {
-            ...emberPlugin.parsers['ember-template-tag'],
+    ...(isEmberPluginAvailable() && {
+        'ember-template-tag': {
+            ...getEmberPlugin().parsers['ember-template-tag'],
             preprocess: emberPreprocessor,
-        };
-    },
+        },
+    }),
     flow: {
         ...flowParsers.flow,
         preprocess: defaultPreprocessor,
     },
-    get oxc() {
-        const oxcPlugin = getOxcPlugin();
-
-        return {
-            ...oxcPlugin.parsers.oxc,
+    ...(isOxcPluginAvailable() && {
+        oxc: {
+            ...getOxcPlugin().parsers.oxc,
             preprocess: defaultPreprocessor,
-        };
-    },
-    get 'oxc-ts'() {
-        const oxcPlugin = getOxcPlugin();
-
-        return {
-            ...oxcPlugin.parsers['oxc-ts'],
+        },
+        'oxc-ts': {
+            ...getOxcPlugin().parsers['oxc-ts'],
             preprocess: defaultPreprocessor,
-        };
-    },
+        },
+    }),
     typescript: {
         ...typescriptParsers.typescript,
         preprocess: defaultPreprocessor,
@@ -127,9 +137,7 @@ export const parsers = {
 };
 
 export const printers = {
-    get 'estree-oxc'() {
-        const oxcPlugin = getOxcPlugin();
-
-        return oxcPlugin.printers['estree-oxc'];
-    },
+    ...(isOxcPluginAvailable() && {
+        'estree-oxc': getOxcPlugin().printers['estree-oxc'],
+    }),
 };


fbartho:
@IanVS — I’m not certain this would work, but an alternative approach might be to use a `Proxy` object instead.

Depending on exactly what they’re doing when they iterate the parsers, maybe that’s sufficient? (A proxy object might delay the activation of the require-expressions until a later phase)

Ideally: testing for existence of the peerDependency would be delayed from GlobalScope until “time of parsing the first file that’s needed”
That might be a bit optimistic, but delaying it until “time of parsing the first file” might be enough.

Unfortunately, I suspect that whoever is iterating the object is probably doing so to wrap every other parser in their own helpers, and that’s a chicken-and-egg unfairness (which plugin gets to register itself last?).
They might not be motivated to revert their code if it keeps some users from seeing a parser failure.

IanVS:
>Unfortunately, I suspect that whoever is iterating the object is probably doing so to wrap every other parser in their own helpers

I don't think that's actually what's happening, I think it's just getting caught up in some option parsing that's converting them to camelCase.  I'm going to revert this PR for now to fix the other issue, and will be happy to review other approaches like the use of Proxy objects to solve the problem reported in https://github.com/IanVS/prettier-plugin-sort-imports/issues/233.

TheCodeDestroyer:
This might be a bit primitive, but this is how I solved issue with prettier require/esm clash for my prettier packages I use this method to resolve the plugins:

```ts
const getPackagePath = (packageName: string): string | null => {
  try {
    if (
      typeof import.meta !== 'undefined' &&
      typeof import.meta.resolve === 'function'
    ) {
      return import.meta.resolve(packageName);
    } else if (
      typeof require !== 'undefined' &&
      typeof require.resolve === 'function'
    ) {
      return require.resolve(packageName);
    }
    throw new Error(
      `Cannot resolve package '${packageName}' in this environment.`,
    );
  } catch (error) {
    console.error(error);
    return null;
  }
};
```

TheCodeDestroyer:
> This might be a bit primitive, but this is how I solved issue with prettier require/esm clash for my prettier packages I use this method to resolve the plugins (maybe it helps):
> 
> ```ts
> const getPackagePath = (packageName: string): string | null => {
>   try {
>     if (
>       typeof import.meta !== 'undefined' &&
>       typeof import.meta.resolve === 'function'
>     ) {
>       return import.meta.resolve(packageName);
>     } else if (
>       typeof require !== 'undefined' &&
>       typeof require.resolve === 'function'
>     ) {
>       return require.resolve(packageName);
>     }
>     throw new Error(
>       `Cannot resolve package '${packageName}' in this environment.`,
>     );
>   } catch (error) {
>     console.error(error);
>     return null;
>   }
> };
> ```

Scratch that... Would maybe this work:
```ts
export const parsers = {
 // parsers without oxc, oxc-ts and ember-template-tag
};

if (isOxcPluginAvailable()) {
  Object.defineProperty(parsers, 'oxc', {
    get() {
      const oxcPlugin = getOxcPlugin();

      return {
        ...oxcPlugin.parsers.oxc,
        preprocess: defaultPreprocessor,
      };
    },
  });
}
```

IanVS:
@TheCodeDestroyer would you like to submit a PR with that approach, and we can test it out?

Issue #242, opened 2025-10-17
Add a custom variable for `subpath imports`


jasikpark:
**Is your feature request related to a problem?**

I'd like to simply order where my [subpath imports](https://nodejs.org/api/packages.html#subpath-imports), `import { abc } from "#xyz"` go in my sorted imports.

**Describe the solution you'd like**

I'd like a keyword for the config like `<BUILTIN_MODULES>`, perhaps `<SUBPATH_IMPORTS>`?

```js
{
importOrder: [
    '<BUILTIN_MODULES>',
    '<THIRD_PARTY_MODULES>',
    '<SUBPATH_IMPORTS>',
    '^(?!.*[.]css$)[./].*$',
    '.css$',
  ],
}
```

Where `<SUBPATH_IMPORTS>` = the regex `'^(#)(.*)$'` or some simplification therein.

**Describe alternatives you've considered**

The current solution works well enough, but has to be updated with new subpath imports:

```js
{
importOrder: [
    '<BUILTIN_MODULES>',
    '<THIRD_PARTY_MODULES>',
    '^(#api|#assets|#components|#hooks|#integrations|#mocks|#storybook|#test-setup|#util)(.*)$',
    '^(?!.*[.]css$)[./].*$',
    '.css$',
  ],
```

**Additional context**

https://nodejs.org/api/packages.html#subpath-imports

fbartho:
Can you not do this? `'^(#.*)(.*)$'`

jasikpark:
Sure can, just seems like a common regex that someone might not come up with originally.

IanVS:
Yep, I agree with @fbartho there.  There's an even simpler pattern which we can suggest that folks use for subpath imports: `'^#.+'`.

At its core, this plugin requires some knowledge of regex to customize.  But that's why we try to give some examples in the README.  I'd welcome a contribution to the docs that mentions subpath imports.

PR #243, opened 2025-10-20
Add note for how to specify subpath import regex to README

jasikpark:
Figured this should go with the existing local alias examples, it's a simpler approach, and presumably the new state-of-art for local aliases, so maybe it should go higher in the examples?

Followup to #242

### DIFF
--- a/README.md
+++ b/README.md
@@ -319,7 +319,7 @@ import App from './App';
 
 ##### 6. Group aliases with local imports
 
-If you define non-relative aliases to refer to local files without long chains of `"../../../"`, you can include those aliases in your `importOrder` to keep them grouped with your local code.
+If you use some other method to define non-relative aliases to refer to local files without long chains of `"../../../"`, you can include those aliases in your `importOrder` to keep them grouped with your local code.  If you use the common `@` symbol for these aliases, you may want some way to group them separately from scoped npm packages, which can be done like this:
 
 ```json
 "importOrder": [


README.md
```diff
@@ -317,6 +317,26 @@ import icon from '@assets/icon';
 import App from './App';
 ```
 
+##### 7. Group subpath import style local aliases
+
+If you define non-relative aliases to refer to local files without long chains of `"../../../"` via subpath imports, you can include those aliases in your `importOrder` to keep them grouped with your local code.
```
IanVS: It might be nice to use this as an opportunity to teach devs about subpath imports as well, maybe with a link to the node.js docs.  And maybe we can reword this to be a bit less copy-paste from the aliases above?   I can take a crack at a "suggestion" if you'd like.

jasikpark: Certainly please provide a suggestion 👍

README.md
```diff
@@ -297,7 +297,27 @@ import { createBrowserHistory } from 'history';
 import App from './App';
 ```
 
-##### 5. Group aliases with local imports
+##### 5. Group subpath import style local aliases
+
+If you define non-relative aliases to refer to local files without long chains of `"../../../"` via [subpath imports](https://nodejs.org/api/packages.html#subpath-imports), you can include those aliases in your `importOrder` to keep them grouped with your local code.
```
IanVS:
```suggestion
If you are using [subpath imports](https://nodejs.org/api/packages.html#subpath-imports) (local imports starting with "#"), you can include those in your `importOrder` to keep them grouped with your local code, perhaps just above your relative imports, as shown below.
```

IanVS (commented):
How's this?

PR #250, opened 2026-03-09
Fix range formatting import boundary whitespace and skip unrelated import sorting

denbezrukov:
> Built with Codex assistance.

## How the fix works

  Use this example:

  ```ts
  import z from "z";
  import a from "a";


  // next block
  const value = 1;
```

  Now imagine only the import area is being range-formatted.

  ### What the plugin wants to output

  After sorting, the desired result is:

  ```ts
  import a from "a";
  import z from "z";


  // next block
  const value = 1;
```

  That means:

  - imports are sorted
  - the blank line after imports is preserved

  ### What was going wrong before

  Inside the plugin, blank-line separators are represented by a synthetic marker node.

  Babel prints that marker as a real top-level line:

  ```ts
  import a from "a";
  import z from "z";
  "PRETTIER_PLUGIN_SORT_IMPORTS_NEW_LINE";
```

  At the same time, the plugin also preserved the original whitespace after the removed import block:

  ```ts
  \n\n// next block
  const value = 1;
```

  When those two pieces were joined together, the separator from the generated import block and the
  preserved whitespace from the original file could stack, producing extra blank lines:

  ```ts
  import a from "a";
  import z from "z";



  // next block
  const value = 1;
```

  That was the original bug.

  ### Commit 1

  This commit fixes the synthetic marker handling.

  Instead of leaving Babel’s printed marker line in place, it replaces each generated marker run with
  exactly one real separator:

  ```ts
  \n\n
```
  So the generated import block becomes:

  ```ts
  import a from "a";
  import z from "z";
```

  That removes one source of duplicate spacing, but it is not enough by itself.

  ### Commit 2

  This commit fixes the join between the rewritten imports and the untouched rest of the file.

  The remainder of the file may already begin with blank lines:

  ```ts
  \n\n// next block
  const value = 1;
```

  If that is appended directly to imports that already end with \n\n, the result still has too much
  space.

  So this commit changes the boundary logic to:

  - keep the separator emitted by the rewritten import block
  - remove only the duplicated overlap from the original remainder
  - preserve any extra spacing that already existed in the original file

  So the final result becomes:

  ```ts
  import a from "a";
  import z from "z";


  // next block
  const value = 1;
```

  ### Commit 3

  This commit changes when import sorting runs during range formatting.

  Example:

  ```ts
  import z from "z";
  import a from "a";

  const value = 1;
  console.log(value);
```
  If the selected range is only:

  ```ts
  console.log(value);
```

  imports should not be touched.

  Before this change, the plugin could still sort imports even when the formatted range did not overlap
  the import block.

  This commit adds overlap gating:

  - if the selected range does not touch imports, skip import sorting
  - if the selected range does touch imports, keep sorting imports as before

  So:

  - formatting unrelated code no longer changes imports
  - formatting imports still sorts them

  ### Commit 4

  This commit fixes a comment edge case.

  Example:

  ```ts
  import z from "z";
  import a from "a";


  // next block
  const value = 1;
```

  Babel may attach // next block to the last import.

  The spacing-preservation logic was measuring the original boundary from too far down, effectively
  from the attached comment instead of from the end of the last import. That caused the plugin to miss
  the original blank lines before the comment.

  This commit changes that behavior to:

  - measure the original boundary from the end of the last import
  - not from an attached comment node

  That makes spacing preservation work correctly for comment-adjacent boundaries too.

  ### Final behavior

  If the selected range overlaps imports:

  Input:

  ```ts
  import z from "z";
  import a from "a";


  // next block
  const value = 1;
```
  Output:
  ```ts
  import a from "a";
  import z from "z";


  // next block
  const value = 1;
```
  If the selected range does not overlap imports:

  ```ts
  const value = 1;
```

  imports remain unchanged.

### DIFF
--- a/src/preprocessors/vue.ts
+++ b/src/preprocessors/vue.ts
@@ -3,6 +3,11 @@ import type { SFCDescriptor } from '@vue/compiler-sfc';
 import { ImportOrderParserPlugin } from '../../types';
 import { PrettierOptions } from '../types';
 import { hasPlugin } from '../utils/get-experimental-parser-plugins';
+import {
+    getPartialFormatRange,
+    toLocalFormatRange,
+    type FormatRange,
+} from './format-range';
 import { preprocessor } from './preprocessor';
 
 // Non-exhaustive, describes just the properties needed for type guarding
@@ -12,6 +17,7 @@ type VueBlock = BaseBlock | LocBlock;
 
 export function vuePreprocessor(code: string, options: PrettierOptions) {
     try {
+        const fileFormatRange = getVueFormatRange(code, options);
         const { parse } = require('@vue/compiler-sfc');
         const version =
             require('@vue/compiler-sfc/package.json').version?.split('.')[0];
@@ -50,7 +56,16 @@ export function vuePreprocessor(code: string, options: PrettierOptions) {
             const end = isLocBlock(block)
                 ? block.loc.end.offset
                 : (block as BaseBlock).end.offset;
-            const preprocessedBlockCode = sortScript(block, options);
+            const preprocessedBlockCode = getPreprocessedBlockCode(
+                block,
+                options,
+                {
+                    start,
+                    end,
+                    fileFormatRange,
+                    originalCode: code,
+                },
+            );
             result += code.slice(offset, start) + preprocessedBlockCode;
             offset = end;
         }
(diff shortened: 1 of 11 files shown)

### CODE REVIEW

File: src/types.ts [hunk 1]
```diff
@@ -13,7 +13,10 @@ import { chunkTypeOther, chunkTypeUnsortable } from './constants';
 import { examineAndNormalizePluginOptions } from './utils/normalize-plugin-options';
 
 export interface PrettierOptions
-    extends Required<PluginConfig>, RequiredOptions {}
+    extends Required<PluginConfig>, RequiredOptions {
+    ppsiVueOriginalRangeStart?: number;
+    ppsiVueOriginalRangeEnd?: number;
```
IanVS, line 18: Where are these coming from?

IanVS (commented):
I'll need to spend some more time reviewing this, but I know in the past we had a hard time determining whether only a range was being formatted.  I'm curious where `ppsiVueOriginalRangeStart` and `ppsiVueOriginalRangeEnd` come from, and why they're not in the standard type, if they're something we can rely on.

src/preprocessors/vue.ts
```diff
@@ -68,6 +83,66 @@ export function vuePreprocessor(code: string, options: PrettierOptions) {
     }
 }
 
+function getVueFormatRange(code: string, options: PrettierOptions) {
+    const currentFormatRange = getPartialFormatRange(code, options);
+
+    if (currentFormatRange) {
+        options.ppsiVueOriginalRangeStart = currentFormatRange.start;
```
denbezrukov, line 90: These aren't standard Prettier options, they're just internal fields we stick on the options object in the Vue preprocessor so we can remember the original file-level range across the extra Vue preprocessing step. That's why they're only in our local type.

ppsi is short for prettier-plugin-sort-imports 😀

IanVS: I'm blind, thanks.  Like I said, I didn't look too hard (obviously).  I may have other questions as I go through this though.  Thanks for tackling this!

IanVS: I haven't forgotten about this.  Work is just a bit crazy at the moment, but I'll try to prioritize a review on this as soon as I can.

